PR #9, opened 2015-10-10
Dedicated php-fpm process

polytan02:
I modified the empty app to very easily let the end app maker use a dedicated php-fpm process.

The base app is not modified, nor its behavior.


opi:
thx

PR #18, opened 2017-02-03
Upgrade source

maniackcrudelis:
L'inclusion des sources directement est proscrite.

Il y a différentes manière d'intégrer les sources.
L'adresse de la source et la somme de contrôle devrait servir dans tout les cas.

### DIFF
--- a/conf/app.src
+++ b/conf/app.src
@@ -1,0 +1,2 @@
+SOURCE_URL=url of app's source
+SOURCE_SUM=md5 or sha1 checksum
--- a/sources/README.md
+++ b/sources/README.md
@@ -1,3 +1,0 @@
-Place the sources of the webapp in this folder.
-
-This file can be removed.
--- a/sources/app.src
+++ b/sources/app.src
@@ -1,2 +1,0 @@
-SOURCE_URL=url of app's source
-SOURCE_SUM=md5 or sha1 checksum
--- a/sources/app.src
+++ b/sources/app.src
@@ -1,0 +1,2 @@
+SOURCE_URL=url of app's source
+SOURCE_SUM=md5 or sha1 checksum


M5oul:
On the linter, I added [a check about how sources are managed](https://github.com/YunoHost/package_linter/commit/2d28219a9a62f90be016e59b30feebff15e700fb).
If `sources` folder exists, it consider it's wrong.
I think this information could be be stored on a file like `conf/sources` as `conf/version` used on some packages.

maniackcrudelis:
Oui pas de problème.
Le nom app.src, que j'ai piqué à ljf ne te convient pas?

Et qu'en est-il des patch? On les placerais aussi dans /conf ?

M5oul:
`app.src` it's ok.
Patch could also be on `conf`.

zamentur:
Note app.src is the src file by default. Because it is possible an app has several sources. In this case the future ynh_setup_sources helpers should be able to manage it too.

opi:
( unrelated to this PR )

> If sources folder exists, it consider it's wrong.

Sometimes it's ok for `sources` folder to exists, at least for some .patches files; Remember that `{conf, scripts, sources}` folders are the only one to be copied during installation, and available during update process. Using another folder name will lead to failing upgrade.

M5oul:
Ok, thanks for this precious information.
We may not use `sources` folder anymore.
But, we may use `conf` folder for conf, patches…etc.

opi:
> We may not use sources folder anymore.
> But, we may use conf folder for conf, patches…etc.

disagree ; patches are source related files, not configuration.

M5oul:
I am agree with you, that's technically not.
But, for the linter, the [easy way to check that sources are stored locally is to check if `sources` folder exists](https://github.com/YunoHost/package_linter/commit/2d28219a9a62f90be016e59b30feebff15e700fb#diff-2cb3a7966fcf81dc02ba8722d74be525R70).

maniackcrudelis:
Maybe you can check if the folder contains another files than somes stricts files and folders. Like patch, app.src and some others like that.

M5oul:
Good idea.

zamentur:
> Sometimes it's ok for sources folder to exists, at least for some .patches files; Remember that {conf, scripts, sources} folders are the only one to be copied during installation, and available during update process. Using another folder name will lead to failing upgrade.

@opi are you sure about that ? nextcloud package has /patches folder and it seems available.
Here is part of the code which seems to do this:
https://github.com/YunoHost/yunohost/blob/unstable/src/yunohost/app.py#L1173

Note: I have seen packages with patches dir at the root of the app, but no patches dir in sources

opi:
@zamentur not perfectly sure ;)

Here is what I have on a production server, for nextcloud : 
```
ls -l /etc/yunohost/apps/nextcloud/
total 16
-r-------- 1 root  root 2150 oct.   9 15:36 manifest.json
dr-------- 2 admin 1007 4096 oct.   9 15:36 scripts
-r-------- 1 root  root  227 oct.   9 15:38 settings.yml
-r-------- 1 root  root  209 oct.   9 15:38 status.json
```
So no `patches` folder (not even `conf` or `sources`, in any installed apps).

Not sure if it's troublesome, but we need to be aware; I first encountered this issue when playing around with ansible (https://github.com/opi/anarchism_ynh/tree/ansible-structured), and the `ansible` folder was not inside the `scripts` folder; Install went fine, but during upgrade, or remove/backup there were some "file not found" issue.

opi:
Now that I wrote about, this is not during the upgrade process, but during remove/backup/restore ; So using patches at these times is not relevant ; Maybe everything I said before was not relevant then ;)

PR #21, opened 2017-02-07
[enh] Update remove

maniackcrudelis:
- Suppression sécurisée de $final_path, ce serait bête un `sudo rm -rf /var/www/`
- Et toujours php-fpm par défaut...

### DIFF
--- a/scripts/remove
+++ b/scripts/remove
@@ -10,14 +10,19 @@ source /usr/share/yunohost/helpers
 domain=$(ynh_app_setting_get "$app" domain)
 
 # Remove sources
-if test -z "$app" || [ "$app" = "/" ]; then
-	echo "Variable app is empty, suppression of app directory cancelled." >&2
+# Avoid removing important directory if $app is empty or equal to 
+# "/" "/var" "/var/www" "/opt" "/home" "/home/yunohost.app"
+src_path=/var/www/$app
+if test -z "$src_path" || [ "/var/www /opt /home/yunohost.app" =~ $src_path ]; then
+	echo "Variable src_path seems incomplete, suppression of app directory cancelled." >&2
 else
-	sudo rm -rf /var/www/$app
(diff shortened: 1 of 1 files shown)


scripts/remove
```diff
@@ -10,15 +10,24 @@ source /usr/share/yunohost/helpers
 domain=$(ynh_app_setting_get "$app" domain)
 
 # Remove sources
-sudo rm -rf /var/www/$app
+# Avoid removing important directory if $app is empty or $src_path equal to 
+# "/" "/var" "/var/www" "/opt" "/home" "/home/yunohost.app"
+src_path=/var/www/$app
+if test -z "$app" || [ "/var/www /opt /home/yunohost.app" =~ $src_path ]; then
+	echo "Variable src_path seems incomplete, suppression of app directory cancelled." >&2
```
JimboJoe: Here the empty `$app` case isn't needed as it can be included in the second test (if you add a trailing `/` to `/var/www`).
Moreover, two small typos : _**un**complete_ and _deletion_.

scripts/remove
```diff
@@ -10,15 +10,24 @@ source /usr/share/yunohost/helpers
...
+else
+	sudo rm -rf $src_path
+fi
 
 # Remove nginx configuration file
 sudo rm -f /etc/nginx/conf.d/$domain.d/$app.conf
+# Reload nginx service
+sudo service nginx reload
```
JimboJoe: I am in favor of systemd syntax (but maybe in another PR, as it needs to be changed in every script) : `systemctl reload nginx`.

scith:
Agreed

scripts/remove
zamentur: Moul is right, it's incomplete

scripts/remove
magikcypress: +1

PR #23, opened 2017-02-07
[enh] Update upgrade script

maniackcrudelis:
- Source distante avec cheksum
- php-fpm...

### DIFF
--- a/scripts/upgrade
+++ b/scripts/upgrade
@@ -19,10 +19,21 @@ language=$(ynh_app_setting_get "$app" language)
 # Remove trailing "/" for next commands
 path_url=${path_url%/}
 
-# Copy source files
+# Obtain and extract source files
+src_url=$(cat ../conf/app.src | grep SOURCE_URL | cut -d= -f2)
+src_checksum=$(cat ../conf/app.src | grep SOURCE_SUM | cut -d= -f2)
 src_path=/var/www/$app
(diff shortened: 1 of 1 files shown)


scripts/upgrade
```diff
@@ -19,10 +19,21 @@ language=$(ynh_app_setting_get "$app" language)
 # Remove trailing "/" for next commands
 path_url=${path_url%/}
 
-# Copy source files
+# Obtain and extract source files
+src_url=$(cat ../conf/app.src | grep SOURCE_URL | cut -d= -f2)
+src_checksum=$(cat ../conf/app.src | grep SOURCE_SUM | cut -d= -f2)
```
zamentur: It could be 
source ../conf/app.src
And next use SOURCE_URL and SOURCE_SUM like var

My own app.src format can contain 
"SUM_PRG=md5sum", because by default I have consider sha256sum
"SOURCE_FILE=exemple.tar.gz" like that we are able to uncompress if it zip or bz2, and we can search sources in "/opt/yunohost-apps-src/$YNH_APP_ID/$SOURCE_FILE" (for offline user cases)

maniackcrudelis: I never thought to use a small conf file like that, but I think it works.

JimboJoe: That's indeed a clever way to load parameters!

PR #29, opened 2017-06-02
[enh] Add extra_files directory

maniackcrudelis:


scith:
What is this directoy for?

maniackcrudelis:
Like `patches`, a directory for ynh_setup_source. This one is used for additional files you want to add to your final_path after the uncompression of the original source.

PR #30, opened 2017-06-02
[Enh] Add patches directory

maniackcrudelis:


scith:
What is this directory for?

maniackcrudelis:
This directory, as the other one extra_files, is a directory uses by ynh_setup_source.
This one is used to contains somes patch files.

More info [here](https://github.com/YunoHost/yunohost/blob/unstable/data/helpers.d/utils#L124-L132)

JimboJoe:
An example of use here: https://github.com/YunoHost-Apps/wallabag2_ynh/tree/master/sources/patches

PR #32, opened 2017-06-02
[enh] Upgrade backup

maniackcrudelis:


maniackcrudelis:
We're currently in version 2.5.6. Please note this PR need a next release.

scripts/backup
```diff
@@ -1,33 +1,81 @@
...
+db_name=$(ynh_app_setting_get $app db_name)
+db_pwd=$(ynh_app_setting_get $app mysqlpwd)
+
+#=================================================
+# STANDARD BACKUP STEPS
+#=================================================
+# BACKUP OF THE MAIN DIR OF THE APP
+#=================================================
+
+ynh_backup "$final_path" "${backup_dir}$final_path"
```
JimboJoe: Wouldn't it possible to get [the same syntax](https://github.com/YunoHost/test_apps/blob/master/backup_recommended_app_ynh/scripts/backup#L21) of the future helper (i.e. not specifying the destination)?

maniackcrudelis: No we can't do that, because before the new helper, the old one will be used.
And this old helper ynh_backup doesn't support this syntax

JimboJoe: I get that, but as we can supersede the ynh_backup helper in the _common.sh, why not?

maniackcrudelis: No we can't.
If we add a ynh_backup in the common, it will be load and replace by the current ynh_backup.
Because, in the script we load first the common, and then the official helpers.
https://github.com/YunoHost/example_ynh/blob/upgrade_backup/scripts/backup#L21-L22

Like that, the official helpers are always the main helpers.

JimboJoe: OK I forgot that order... I still feel a bit unsatisfied, but I guess I'll have to live with that :-D

maniackcrudelis: It's just temporary you know, so until our next release. Even if it seems so far ;)
I hope the next week, we'll do a meeting to talk about that.

But, for now, it works with the old helper, and with the new one.

PR #35, opened 2017-06-02
[enh] Upgrade upgrade script

maniackcrudelis:


maniackcrudelis:
We're currently in version 2.5.6. Please note this PR need a next release.
Need also https://github.com/YunoHost/yunohost/pull/284, https://github.com/YunoHost/yunohost/pull/285 and https://github.com/YunoHost/yunohost/pull/287

maniackcrudelis:
> OK but perhaps we could put the path normalization under "Fix old things"?

Why ? This is a normalization of the path, as it gives by the user. So I think we always have to use it.

PR #39, opened 2017-07-05
Remove sudo prefixes (not needed any more in YunoHost 2.6)

JimboJoe:
## Problem
Sudo is no more necessary. If we forgot one it could be ok in 2.6.4 but not in 2.5.

## Solution
We encourage to put a requirements to yunohost 2.6.4 and remove sudo from this example app.

## PR Status
Work finished. Could be reviewed and tested.

## Validation

- [x] **Principle agreement** 2/2 : ljf, Maniack C
- [x] **Quick review** 1/1 : ljf
- [ ] **Simple test** 0/1 : 
- [x] **Deep review** 1/1: Maniack C
- [ ] **Complex test** 0/0

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -14,7 +14,7 @@
         "url": "http://example.com"
     },
     "requirements": {
-        "yunohost": ">> 2.6.4"
+        "yunohost": ">= 2.6.4"
     },
     "multi_instance": true,
     "services": [
--- a/manifest.json
+++ b/manifest.json
@@ -14,7 +14,7 @@
         "url": "http://example.com"
     },
     "requirements": {
-        "yunohost": ">> 2.5.6"
+        "yunohost": ">> 2.6.4"
     },
     "multi_instance": true,
     "services": [
--- a/scripts/restore
+++ b/scripts/restore
@@ -15,8 +15,8 @@ set -eu
 
 if [ ! -e _common.sh ]; then
 	# Get the _common.sh file if it's not in the current directory
-	sudo cp ../settings/scripts/_common.sh ./_common.sh
-	sudo chmod a+rx _common.sh
+	cp ../settings/scripts/_common.sh ./_common.sh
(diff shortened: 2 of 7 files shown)


scripts/restore
```diff
@@ -15,8 +15,8 @@ set -eu
 
 if [ ! -e _common.sh ]; then
 	# Get the _common.sh file if it's not in the current directory
-	sudo cp ../settings/scripts/_common.sh ./_common.sh
-	sudo chmod a+rx _common.sh
+	cp ../settings/scripts/_common.sh ./_common.sh
+	chmod a+rx _common.sh
```
zamentur: Did you test this ?
I think ./ is in readonly mode no?

zamentur: But it's not to do in this PR

maniackcrudelis: So far, it works for me.
Did you change anything recently ?

maniackcrudelis:
How do you expect to try this PR ?
This package doesn't work in state.

manifest.json
```diff
@@ -14,7 +14,7 @@
         "url": "http://example.com"
     },
     "requirements": {
-        "yunohost": ">> 2.5.6"
+        "yunohost": ">> 2.6.4"
```
JimboJoe: Shouldn't we set that as ">="? That is far clearer for me.

JimboJoe: @maniackcrudelis @zamentur, your opinion? :-)

maniackcrudelis: Do we have any documentation about this syntax ?

JimboJoe: You can find it [here](https://github.com/YunoHost/yunohost/blob/unstable/src/yunohost/utils/packages.py#L71-L79).

maniackcrudelis: Yes you're right. As we are in 2.6.4 currently, we have to set at >=

PR #45, opened 2017-12-29
Fix the issue alias_traversal

maniackcrudelis:
https://github.com/yandex/gixy/blob/master/docs/en/plugins/aliastraversal.md

### DIFF
--- a/scripts/change_url
+++ b/scripts/change_url
@@ -59,13 +59,18 @@ if [ $change_path -eq 1 ]
 then
   # Make a backup of the original nginx config file if modified
   ynh_backup_if_checksum_is_different "$nginx_conf_path"
-  # Replace locations starting with old_path
-  # Look for every location possible patterns (see https://nginx.org/en/docs/http/ngx_http_core_module.html#location)
-  ynh_replace_string "location\( \(=\|~\|~\*\|\^~\)\)\? $old_path" "location\1 $new_path"  "$nginx_conf_path"
-  # Replace path in "return" directives
-  ynh_replace_string "return \([[:digit:]]\{3\}\) $old_path" "return \1 $new_path"  "$nginx_conf_path"
-  # Calculate and store the nginx config file checksum
-  ynh_store_file_checksum "$nginx_conf_path"
+	# Set global variables for nginx helper
+	domain="$old_domain"
+	path_url="$new_path"
+	# Store path_url setting
+	ynh_app_setting_set $app path_url "$path_url"
+	# Create a dedicated nginx config
+	ynh_add_nginx_config
+	if [ "$path_url" != "/" ]
+	then
+		ynh_replace_string "^#sub_path_only" "" "$nginx_conf_path"
+	fi
+	ynh_store_file_checksum "$nginx_conf_path"
 fi
 
 # Change the domain for nginx
--- a/scripts/change_url
+++ b/scripts/change_url
@@ -66,6 +66,11 @@ then
 	ynh_app_setting_set $app path_url "$path_url"
 	# Create a dedicated nginx config
 	ynh_add_nginx_config
+	if [ "$path_url" != "/" ]
+	then
+		ynh_replace_string "^#sub_path_only" "" "$nginx_conf_path"
+	fi
+	ynh_store_file_checksum "$nginx_conf_path"
 fi
 
 # Change the domain for nginx
--- a/scripts/change_url
+++ b/scripts/change_url
@@ -66,11 +66,6 @@ then
 	ynh_app_setting_set $app path_url "$path_url"
 	# Create a dedicated nginx config
 	ynh_add_nginx_config
-	if [ "$path_url" != "/" ]
-	then
-		ynh_replace_string "^#sub_path_only" "" "$nginx_conf_path"
-	fi
-	ynh_store_file_checksum "$nginx_conf_path"
 fi
 
 # Change the domain for nginx
--- a/scripts/change_url
+++ b/scripts/change_url
@@ -57,8 +57,8 @@ nginx_conf_path=/etc/nginx/conf.d/$old_domain.d/$app.conf
 # Change the path in the nginx config file
 if [ $change_path -eq 1 ]
 then
-  # Make a backup of the original nginx config file if modified
-  ynh_backup_if_checksum_is_different "$nginx_conf_path"
+	# Make a backup of the original nginx config file if modified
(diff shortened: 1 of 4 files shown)


Josue-T:
Hello,

Does your patch support the change-url script for all case ?
For example : 
```
/oldpath/ -> /newpath/
/ -> /newpath/
/oldpath/ -> /
/ -> / (other domain) 
...
```

maniackcrudelis:
> Does your patch support the change-url script for all case ?

Not at all...
Here, it's only for install and upgrade. To handle that in a change_url script, you can use this:
```
# Move from sub path to root
if [ "$new_path" == "/" ]
then
	ynh_replace_string "\(^.*rewrite.*\^$old_path.* permanent;\)" "#sub_path_only\1" "$nginx_conf_path"
	ynh_replace_string "\(rewrite *\^\)$old_path\$ $old_path/*" "\1$new_path$ $new_path" "$nginx_conf_path"

# Move to a sub path
else
	ynh_replace_string "^#sub_path_only" "" "$nginx_conf_path"
	ynh_replace_string "\(rewrite *\^\)$old_path\$ $old_path/*" "\1$new_path$ $new_path/" "$nginx_conf_path"
fi
```

JimboJoe:
Here I think we could now depend on YunoHost 2.7.7 and [this PR](https://github.com/YunoHost/yunohost/pull/407) to change the way `change_url` works (by using the original `nginx.conf`), couldn't we?

Josue-T:
I think we might need to adapt the nginx helper to use the path to the nginx config while we call the change_url script. If the change_url script call `ynh_add_nginx_config` the nginx config path might need to be `../settings/conf/nginx.conf` insted of  `../conf/nginx.conf`

Or does we do a new helper for that ?

JimboJoe:
I'm not sure to understand how it affects the relative path.
I have:
```
# ll /etc/yunohost/apps/piwigo/
total 20
drwxr-xr-x 2 root  root 4096 Jan 24 15:47 conf
-rwxrwxrwx 1 admin 1007 2091 Jan 20 16:23 manifest.json
drwxrwxrwx 2 admin 1007 4096 Jan 24 15:46 scripts
-r-------- 1 root  root  905 Jan 24 15:47 settings.yml
-r-------- 1 root  root  119 Jan 24 15:47 status.json
```
Isn't the relative path from the script to the conf file the same?

Josue-T:
Hello,

I say that because when we call the backup / restore script the path to the other script (like _common.sh) has an other path than when we call the install or upgrade script.

I have made a check and actually the script directory is copied in `/var/cache/yunohost/from_file/scripts` before to be called in this directory. But the conf dir is not. We can see that here : https://github.com/YunoHost/yunohost/blob/unstable/src/yunohost/app.py#L485-L486

So I think you are true to say that we don't need to adapt the nginx helper but in this case we need to adapt the python script because actually in the conf dir we have the file from the last app install, so it's not good.

JimboJoe:
@Josue-T , I totally agree with you and we should certainly make PR on the core for the next stable release...
Is this something we would also need for backup/restore (I doubt it)?

Josue-T:
> Is this something we would also need for backup/restore (I doubt it)?

I think not because the script is called in an other folder and the utilisation is also different.

JimboJoe:
Just proposed a core PR for this [here](https://github.com/YunoHost/yunohost/pull/427).

Jibec:
why don't we merge this?

maniackcrudelis:
Because nobody from the apps group has approved yet.

JimboJoe:
Shall we handle the `change_url` script in this PR, or in a separate one?

maniackcrudelis:
Your PR about change_url has been merged https://github.com/YunoHost/yunohost/pull/427, so I think the better thing to do would be to rewrite the change_url according to the new PR.
And, by the way, to manage alias_traversal.

Josue-T:
After https://github.com/YunoHost/yunohost/pull/462 we could also update with the new version of `ynh_add_nginx_config`

JimboJoe:
Should be nearer to a merge decision... :wink: 
Implemented here: https://github.com/YunoHost-Apps/jirafeau_ynh/pull/47 and https://github.com/YunoHost-Apps/ttrss_ynh/pull/53

Remaining drawbacks in the change_url script:
- the need to set the `path_url` setting otherwise you get a bad behavior on next upgrade...
- the need to deal with nginx conf file checksum

These two points should ideally be obfuscated on the long term...

scripts/change_url
```diff
@@ -59,13 +59,18 @@ if [ $change_path -eq 1 ]
...
   ynh_backup_if_checksum_is_different "$nginx_conf_path"
-  # Replace locations starting with old_path
-  # Look for every location possible patterns (see https://nginx.org/en/docs/http/ngx_http_core_module.html#location)
-  ynh_replace_string "location\( \(=\|~\|~\*\|\^~\)\)\? $old_path" "location\1 $new_path"  "$nginx_conf_path"
-  # Replace path in "return" directives
-  ynh_replace_string "return \([[:digit:]]\{3\}\) $old_path" "return \1 $new_path"  "$nginx_conf_path"
-  # Calculate and store the nginx config file checksum
-  ynh_store_file_checksum "$nginx_conf_path"
+	# Set global variables for nginx helper
+	domain="$old_domain"
```
frju365: Je vais être très énervant(comme d'habitude), mais pourrait-on avoir une indentation correcte :wink: ? C'est une question de simplicité et d'organisation du code.

scripts/install
```diff
@@ -105,6 +105,11 @@ ynh_setup_source "$final_path"
 
 # Create a dedicated nginx config
 ynh_add_nginx_config
+if [ "$path_url" != "/" ]
+then
+	ynh_replace_string "^#sub_path_only" "" "/etc/nginx/conf.d/$domain.d/$app.conf"
```
frju365: "ynh_replace_string "^#sub_path_only " "" "/etc/nginx/conf.d/$domain.d/$app.conf"" --> is better to avoid useless indentation ;-)
for Jirafeau after upgrade, I have : 
``
 rewrite ^/jirafeau$ /jirafeau/ permanent;
location /jirafeau/ {
```

JimboJoe:
I've taken your remarks into account, and I encourage you to make these kind of changes directly by yourself next time... :innocent:

zamentur:
As it's a security issue and the jinja discussion seems difficult. I merge this PR.

PR #46, opened 2018-01-17
use ynh_add_systemd_config and doublequotes

Jibec:


Jibec:
kind remember

maniackcrudelis:
Is it really usefull to add so many quotes everywhere ?
$path_url can't contains any special characters or spaces.
Neither do $domain or $app, same thing for $admin and $is_public.

Maybe $language can have a space.

Again, $port is a numeric value, that's all.
$db_name is from $app, so...

$final_path, maybe, if the variable isn't set with /var/www/$app.

Gofannon:
1. Is it possible to fix the typo "ynh_add_systemd_config" in the first place please? I made the mistake because it was in the example :/
2. About the quotes, having "best practices" should be interesting to avoid having to thing about it
    - they are needed for the helper `ynh_replace_string` at least :)
3. Is this the right place to discuss about this?

maniackcrudelis:
Done...
https://github.com/YunoHost/example_ynh/commit/f24034fcfc8f4a4dc3a4ab4c10fd8c1017c158c5

zamentur:
$path_url can't contains any special characters or spaces.

path_url could contains space, is there a check to avoid that ?

maniackcrudelis:
If we suppose that the core allow a path_url with a space, it's not going to go further than `ynh_normalize_url_path` that takes only $1 as path_url variable.
Never tried though.

Jibec:
Remember of our old discussion: bash scripting is unknown to most. Adding quotes is a good practice for newbies. That's also the suggestion of shellcheck.

It is obvious it can be useless for many variables.
I'm fine with no merge at all.

zamentur:
I suggest to close this PR

Jibec:
I'm fine with this PR being closed

PR #47, opened 2018-06-28
Add comments

maniackcrudelis:
I added many comments to explain what is used for what. What can be removed and others information.
I also updated the readme and some other parts of the scripts.

I don't know why there's the previous commits, since à git pull before starting to work. Never mind...

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,11 @@
+# Usage of this package (REMOVE THIS SECTION BEFORE RELEASE)
+- Copy this app before working on it.
+- Edit `conf/nginx.conf` file to match application prerequisites.
+- Edit `manifest.json` with application specific information.
+- Edit the `install`, `upgrade`, `remove`, `backup`, and `restore` scripts.
+- Add a `LICENSE` file for the package.
+- Edit `README.md`.
+
 # Example app for YunoHost
 
 [![Integration level](https://dash.yunohost.org/integration/REPLACEBYYOURAPP.svg)](https://ci-apps.yunohost.org/jenkins/job/REPLACEBYYOURAPP%20%28Community%29/lastBuild/consoleFull)  
@@ -21,7 +29,7 @@ Quick description of this app.
 
 ## Configuration
 
-How to configure this app ? By an admin panel, an plain file with ssh, or any other way.
+How to configure this app: by an admin panel, a plain file with ssh, or any other way.
 
 ## Documentation
 
@@ -32,8 +40,8 @@ How to configure this app ? By an admin panel, an plain file with ssh, or any ot
 
 #### Multi-users support
 
-Is LDAP and HTTP auth supported ?
-Can the app be used by multiple users ?
+Are LDAP and HTTP auth supported?
+Can the app be used by multiple users?
 
 #### Supported architectures
 
@@ -45,17 +53,9 @@ Can the app be used by multiple users ?
 
 * Any known limitations.
 
-## Additionnal informations
+## Additional informations
 
 * Other informations you would add about this application
-
-### Usage of this package
-- Copy this app before working on it.
-- Edit `conf/nginx.conf` file to match application prerequisites.
-- Edit `manifest.json` with application specific information.
-- Edit the `install`, `upgrade`, `remove`, `backup`, and `restore` scripts.
-- Add a `LICENSE` file for the package.
-- Edit `README.md`.
 
 **More information on the documentation page:**  
 https://yunohost.org/packaging_apps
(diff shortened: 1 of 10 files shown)


JimboJoe:
Great work!
I've just pushed some changes: mostly typo fixes, and set the `port` variable retrieval back in the `remove` script (it's used).

README.md
```diff
@@ -1,22 +1,82 @@
...
+#### Multi-users support
+
+Are LDAP and HTTP auth supported?
+Can the app be used by multiple users?
+
+#### Supported architectures
+
+* x86-64b - [![Build Status](https://ci-apps.yunohost.org/jenkins/job/REPLACEBYYOURAPP%20(Community)/badge/icon)](https://ci-apps.yunohost.org/jenkins/job/REPLACEBYYOURAPP%20(Community)/)
+* ARMv8-A - [![Build Status](https://ci-apps-arm.yunohost.org/jenkins/job/REPLACEBYYOURAPP%20(Community)%20(%7EARM%7E)/badge/icon)](https://ci-apps-arm.yunohost.org/jenkins/job/REPLACEBYYOURAPP%20(Community)%20(%7EARM%7E)/)
+* Jessie x86-64b - [![Build Status](https://ci-stretch.nohost.me/jenkins/job/REPLACEBYYOURAPP%20(Community)/badge/icon)](https://ci-stretch.nohost.me/jenkins/job/REPLACEBYYOURAPP%20(Community)/)
```
JimboJoe: Should we really keep that support here?

scripts/change_url
```diff
@@ -66,11 +66,6 @@ then
 	ynh_app_setting_set $app path_url "$path_url"
 	# Create a dedicated nginx config
 	ynh_add_nginx_config
-	if [ "$path_url" != "/" ]
-	then
-		ynh_replace_string "^#sub_path_only " "" "$nginx_conf_path"
-	fi
-	ynh_store_file_checksum "$nginx_conf_path"
```
JimboJoe: Why removing these lines here (they are still in the install script)?

manifest.json
```diff
@@ -15,7 +15,7 @@
         "url": "http://example.com"
     },
     "requirements": {
-        "yunohost": ">= 2.7.2"
+        "yunohost": ">= 2.7.14"
```
JimboJoe: Shouldn't we point to the latest stable version (i.e. 3.0.0)?

README.md
```diff
@@ -45,25 +53,17 @@ Can the app be used by multiple users ?
...
-- Edit `README.md`.
-
 **More information on the documentation page:**  
 https://yunohost.org/packaging_apps
 
 ## Links
 
  * Report a bug: https://github.com/YunoHost-Apps/REPLACEBYYOURAPP_ynh/issues
- * APP website: Link to the official website of this app
+ * App website: Link to the official website of this app
```
maniackcrudelis: APP was supposed to be replaced by the name of the app. But actually, there's no more other occurrences of 'APP', So let's keep 'App', and each packagers will do as he wants.

README.md
maniackcrudelis: It seems interesting to know if an app is still working on jessie. Since jessie will be still used by many users.
We can't just expect that every users will migrate their server, it's not just a simple upgrade.

scripts/change_url
maniackcrudelis: I was quite sure that I removed this also in the install script. We don't need that anymore, it's included in the nginx helper.

manifest.json
maniackcrudelis: I don't think so, If an app can work both on jessie and stretch, we shouldn't force users to migrate.

Josue-T:
Following the discusion [here](https://github.com/YunoHost-Apps/ffsync_ynh/pull/6#discussion_r198734372), I purpose to add some comment about how we define the `final_path`. Actually it defined as `/var/www/$app`, but in some case it's `/opt/yunohost/$app`. So in my point of view it could be good to add a comment like that : 

> In case of an app wich use directly nginx wich is written in html, php the final path is "/var/www/$app"
> For the app which has an internal webserver or which use uwsgi wich is written in python, java, go, ... the final path is "/opt/yunohost/$app".

maniackcrudelis:
Sounds good to me.
Go on

scripts/change_url
JimboJoe: OK for me. Except this last point, for me we're good to go!

maniackcrudelis: Sorry, my bad, I forgot to fix it. Now it's done.

maniackcrudelis:
Let's merge

PR #50, opened 2018-07-16
[fix] Fix "change_url"

Gofannon:
## Problem
- `example_ynh` did not work while working on https://github.com/YunoHost-Apps/dokuwiki_ynh/pull/37

## Solution
- Correct the script to make it run

## PR Status
- [X] Code finished.
- [X] Tested with Package_check.
- [X] Fix or enhancement tested.
- [X] Upgrade from last version tested.
- [X] Can be reviewed and tested.

## Validation
---
*Minor decision*
- **Upgrade previous version** : 
- [x] **Code review** : Maniack C
- [x] **Approval (LGTM)** : Maniack C
- [ ] **Approval (LGTM)** : 
- **CI succeeded** :  
When the PR is marked as ready to merge, you have to wait for 3 days before really merging it.

### DIFF
--- a/scripts/change_url
+++ b/scripts/change_url
@@ -22,6 +22,17 @@ new_path=$YNH_APP_NEW_PATH
 app=$YNH_APP_INSTANCE_NAME
 
 #=================================================
+# LOAD SETTINGS
+#=================================================
+
+# Needed for helper "ynh_add_nginx_config"
+final_path=$(ynh_app_setting_get $app final_path)
+
+# Add settings here as needed by your application
+#db_name=$(ynh_app_setting_get "$app" db_name)
+#db_pwd=$(ynh_app_setting_get $app db_pwd)
+
+#=================================================
 # CHECK THE SYNTAX OF THE PATHS
 #=================================================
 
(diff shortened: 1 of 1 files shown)


Josue-T:
Ok for me.

maniackcrudelis:
Will be merged in 3 days

scripts/change_url
```diff
@@ -63,7 +74,7 @@ then
 	domain="$old_domain"
 	path_url="$new_path"
 	# Store path_url setting
-	ynh_app_setting_set $app path_url "$path_url"
+	ynh_app_setting_set $app path "$path_url"
```
JimboJoe: Good catch, we don't need to set `path_url`!
Yet, now I doubt that we need to set `path`, as it is set automatically by moulinette (see [here](https://github.com/YunoHost/yunohost/blob/unstable/src/yunohost/app.py#L512), or the [original change_url script](https://github.com/YunoHost/example_ynh/commit/1ff00da69ae4de6f227dac8552d4e3535fc49099#diff-16b8ff8f9c56e393f3ef1956d456e92a)), and considering we don't set `domain`.
Keeping it won't hurt, but then we should also change the comment accordingly :wink:

maniackcrudelis: As we're on the example app, let's fix that before merging then.

PR #58, opened 2019-01-25
Add ynh_print_info examples

alexAubin:
Hellow,

so far, I think very few apps use the new `ynh_print_info` helpers. This is a bit sad considering that is significantly improve the user experience :wink: . More precisely, for instance when installing Nextcloud from the webadmin, there is close to *no* feedback (except for a couple warnings) hence the user will get frustrated because of not being sure that the install process is effectively running... 

(all this being related to the fact that before, DEBUG-level output was enabled but was too technical, hence the introduction of `ynh_print_info`)

Pinging @YunoHost/apps since this is relevant for all app packagers : please if you have some time, **take a few minutes to add this kind of lines in your app scripts explaining what is going on, that will make users happier** :stuck_out_tongue_winking_eye:

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -115,7 +115,7 @@ ynh_install_app_dependencies deb1 deb2
 ###		- Remove also the section "REMOVE THE MYSQL DATABASE" in the remove script
 ###		- As well as the section "RESTORE THE MYSQL DATABASE" in the restore script
 
-ynh_print_info "Ininitializing database ..."
+ynh_print_info "Initializing database ..."
 db_name=$(ynh_sanitize_dbid $app)
 ynh_app_setting_set $app db_name $db_name
 ynh_mysql_setup_db $db_name $db_name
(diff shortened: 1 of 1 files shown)


maniackcrudelis:
That's a pretty good idea, I always wanted to do that. And now that we finally could. I forgot it... :(

maniackcrudelis:
> <kayou[m]> with `ynh_print_info` is it possible to display a progress bar (like wget)?

Now it is https://github.com/YunoHost-Apps/Experimental_helpers/blob/master/ynh_script_progression/ynh_script_progression :grin:

alexAubin:
smol bump

Note that I propose to add a check in package linter for this : https://github.com/YunoHost/package_linter/pull/46

maniackcrudelis:
I'm going to work on a global upgrade of this package this afternoon, meanwhile, let's merge that.
I'll add ynh_print_info in other scripts

PR #60, opened 2019-01-28
[fix] Update example with the new version of php

kay0u:


maniackcrudelis:
LGTM, but that means we'll definitively give up jessie.

PR #64, opened 2019-01-31
Adding Testing and Unstable build

yalh76:


maniackcrudelis:
Not sure that's really interesting to have those links in every apps.
Not even in any apps actually.

Those 2 CI are more for an internal purpose.
Still, we can have other point of views.

yalh76:
Not needed information in the readme

PR #66, opened 2019-02-07
Database actions order

yalh76:
When using other database system than MYSQL, the steps has to be reorder:

- remove script: first remove the database and then remove the dependencies
- restore script: first restore the dependencies and then restore the database

This reorder has no impact for packageq using MYSQL but will help maintainers with other DATABASE systems.


maniackcrudelis:
Good idea indeed !

PR #68, opened 2019-02-10
Global upgrade

maniackcrudelis:
Not that much changes, comparing to what I was expecting...
The next stable release will be more interesting, with many new helpers.

Feel free to modify any text in `ynh_print_info`


scripts/_common.sh
```diff
@@ -1,13 +1,13 @@
...
-#
-# usage: ynh_remove_file_checksum file
-# | arg: file - The file for which the checksum will be deleted
-ynh_delete_file_checksum () {
-	local checksum_setting_name=checksum_${1//[\/ ]/_}	# Replace all '/' and ' ' by '_'
-	ynh_app_setting_delete $app $checksum_setting_name
-}
\ No newline at end of file
+#=================================================
+# PERSONNAL HELPERS
```
JimboJoe:
```suggestion
# PERSONAL HELPERS
```

scripts/_common.sh
```diff
@@ -1,13 +1,13 @@
...
+#=================================================
+
+#=================================================
+# EXPERIMENTAL HELPERS
+#=================================================
+
+#=================================================
+# FUTUR OFFICIAL HELPERS
```
JimboJoe:
```suggestion
# FUTURE OFFICIAL HELPERS
```

scripts/change_url
```diff
@@ -12,6 +12,7 @@ source /usr/share/yunohost/helpers
 #=================================================
 # RETRIEVE ARGUMENTS
 #=================================================
+ynh_print_info "Retrieve arguments from the manifest"
```
JimboJoe: Maybe already a bit too technical :slightly_smiling_face: 
And maybe redundant with next log message ("Load settings")

scripts/change_url
```diff
@@ -62,6 +65,7 @@ fi
 #=================================================
 # MODIFY URL IN NGINX CONF
 #=================================================
+ynh_print_info "Modify url in nginx configuration"
```
JimboJoe: If the idea is to be understandable, shouldn't we precise nginx **web server** configuration...? :thinking:

scripts/install
```diff
@@ -75,22 +76,23 @@ ynh_app_setting_set $app language $language
 #=================================================
 # FIND AND OPEN A PORT
 #=================================================
+ynh_print_info "Configuring firewall ..."
```
JimboJoe: We need to harmonize the tense of verbs: should it be "Configure" of "Configuring"? As it introduces the action, I vote for the latter ("ing" everywhere).

scripts/install
```diff
@@ -75,22 +76,23 @@ ynh_app_setting_set $app language $language
...
 ### `ynh_find_port` will find the first available port starting from the given port.
 ### If you're not using these lines:
 ###		- Remove the section "CLOSE A PORT" in the remove script
 
-ynh_print_info "Configuring firewall ..."
 # Find a free port
 port=$(ynh_find_port 8095)
 # Open this port
-yunohost firewall allow --no-upnp TCP $port 2>&1
+ynh_exec_warn_less yunohost firewall allow --no-upnp TCP $port
```
JimboJoe: Maybe comment on why we're using the `ynh_exec_warn_less` helper here...?

scripts/upgrade
```diff
@@ -26,6 +27,7 @@ db_name=$(ynh_app_setting_get $app db_name)
 #=================================================
 # ENSURE DOWNWARD COMPATIBILITY
 #=================================================
+ynh_print_info "Ensure downward compatibility"
```
JimboJoe: I think this is not really important (and not that clear for the users)...

scripts/upgrade
```diff
@@ -51,6 +53,7 @@ fi
 #=================================================
 # BACKUP BEFORE UPGRADE THEN ACTIVE TRAP
 #=================================================
+ynh_print_info "Backup the app before upgrading"
```
JimboJoe:
```suggestion
ynh_print_info "Backup the app before upgrading (may take a while)"
```

scripts/backup
```diff
@@ -75,3 +82,9 @@ ynh_backup "/etc/systemd/system/$app.service"
 #=================================================
 
 ynh_backup "/etc/cron.d/$app"
+
+#=================================================
+# END OF SCRIPT
+#=================================================
+
+ynh_print_info "Backup script completed. Please wait for YunoHost to create the backup."
```
alexAubin: Hm I'd rather have this directly in the core ... I know it's core's fault and this looks like an immediate solution, but we should avoid every app implementing this themselves : because some of them which at not so maintained won't do it, *and* the core benefits from i18n

scripts/change_url
```diff
@@ -98,5 +102,12 @@ fi
...
 #=================================================
+ynh_print_info "Reload nginx"
 
 systemctl reload nginx
+
+#=================================================
+# END OF SCRIPT
+#=================================================
+
+ynh_print_info "Change of url completed"
```
alexAubin: (Same here, should be in the core I think :S)

scripts/install
```diff
@@ -305,6 +307,12 @@ fi
...
-
 ynh_print_info "Reloading nginx ..."
+
 systemctl reload nginx
+
+#=================================================
+# END OF SCRIPT
+#=================================================
+
+ynh_print_info "Installation completed"
```
alexAubin: (Same here, should be in the core I think :S)

scripts/change_url
maniackcrudelis: Probably clearer indeed.

scripts/install
maniackcrudelis: As you prefer

scripts/install
maniackcrudelis: You mean why we don't want to have stdout ? Or why we were using 2>&1 ?

scripts/upgrade
maniackcrudelis: Still it can takes some times

scripts/upgrade
maniackcrudelis: This will be replaced anyway by the helper with the progression bar later

scripts/backup
maniackcrudelis: That's just a printer. Nothing huge.

scripts/change_url
maniackcrudelis: That's used by the helper with the progression bar actually.
To define the execution time and the progression, we need to have a final execution of the helper.

scripts/install
maniackcrudelis: Same as well

maniackcrudelis:
Bump

If you want to modify some texts, please just do it.
I need this PR to be merged in order to update all Official apps.

scripts/backup
```diff
@@ -87,4 +87,4 @@ ynh_backup "/etc/cron.d/$app"
 # END OF SCRIPT
 #=================================================
 
-ynh_print_info "Backup script completed. Please wait for YunoHost to create the backup."
+ynh_print_info "Backup script completed for $app. (YunoHost will then actually copy those to the archive)."
```
maniackcrudelis: "those" what ?

scripts/backup
```diff
@@ -35,24 +36,28 @@ db_name=$(ynh_app_setting_get $app db_name)
...
+ynh_print_info "Backing up nginx configuration..."
 
 ynh_backup "/etc/nginx/conf.d/$domain.d/$app.conf"
 
 #=================================================
 # BACKUP THE PHP-FPM CONFIGURATION
 #=================================================
+ynh_print_info "Backing up php-fpm configuration..."
 
 ynh_backup "/etc/php/7.0/fpm/pool.d/$app.conf"
```
alexAubin: So, uh, can't we just have block like 

```bash
# BACKUP CONFIGURATION FILES
ynh_print_info "Backing configuration files..."

# Nginx
ynh_backup "/etc/nginx/conf.d/$domain.d/$app.conf"

# php-fpm
ynh_backup "/etc/php/7.0/fpm/pool.d/$app.conf"

# logrotate
ynh_backup "/etc/logrotate.d/$app"

# systemd
ynh_backup "/etc/systemd/system/$app.service"

# cron job
ynh_backup "/etc/cron.d/$app"
```

It looks a bit weird to me to have 5 huge blocks just to have so many similar "small" instructions (those literally just backup one file each time) with the mysql backup (which actually does takes some time) in the middle ...

maniackcrudelis: Hum, I actually prefer the current way because it clearer when you're looking for a part of the backup in the script.
But that's not really important.
Also, I thought, when writing those lines, that it could be interesting for the final user to know what exactly will be backup.

alexAubin:
Applied a few tweaks (I removed a few `ynh_print_info` which, like JimboJoe, I think are really too technical / not a "meaningful step"...)

Feel free to discuss ¯\\\_(ツ)\_/¯

maniackcrudelis:
Ok for me.

scripts/backup
alexAubin: Meh but that's just way too low level ... I mean, in the spirit of YunoHost, maybe a tech-savvy user will know what nginx and php-fpm will refer to, but a regular user shouldn't need to know what those are ...

Imho the purpose of those messages are essentially to let the user know "what the machine is doing right now that is taking so much time". But displaying "5 different messages with technical terms for stuff that lasted basically 0 seconds" looks more like debug than info ... sure it might help a technical user in knowing what's happening under the hood, but it tends to frustrate non-technical user and (from my experience at least) when seeing this they will say "this is so complicated, I have no idea what those mean".

maniackcrudelis: As usual, I'm more in favor of using real name for things than to hide what's really going on.
So I'd rather prefer to add those message to explain what's going to be backup instead of thinking that it's too complicated, so we should hide that from our users.
If an user doesn't understand what is it, he will simply ignore it. For other, It's really useful to know what's going on.

maniackcrudelis:
Do we agree on the last changes about ynh_print_info ?

PR #69, opened 2019-02-10
Spacing

yalh76:


maniackcrudelis:
I did the opposite [here](https://github.com/YunoHost/example_ynh/pull/68/files#diff-ad962c6fd65b725b71aa46c8a715127f)

yalh76:
Ok, I don't know what's the rule...
As far I've seen, after a title, like:
```
# LOAD SETTINGS
```
I've alway seen spacing before code

maniackcrudelis:
Yep, I know, I added `ynh_print_info` without a new line because this line is more part of the title than of the code.
Anyway, there's no real rules about that.

Anyway, you should work from the branch [global_upgrade](https://github.com/YunoHost/example_ynh/tree/global_upgrade) since I've changed some parts in my PR.

PR #70, opened 2019-02-10
remove example to keep something more generic

yalh76:


maniackcrudelis:
Same here, you should work from the branch [global_upgrade](https://github.com/YunoHost/example_ynh/tree/global_upgrade). I've changed the text there.

PR #71, opened 2019-02-10
remove $db_user from remove script.

yalh76:
Not used in other scripts

### DIFF
--- a/scripts/remove
+++ b/scripts/remove
@@ -19,7 +19,6 @@ app=$YNH_APP_INSTANCE_NAME
 domain=$(ynh_app_setting_get $app domain)
 port=$(ynh_app_setting_get $app port)
 db_name=$(ynh_app_setting_get $app db_name)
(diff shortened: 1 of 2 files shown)


yalh76:
and removed ynh_script_progression: not a common helper

scripts/remove
```diff
@@ -49,7 +48,7 @@ ynh_remove_systemd_config
 ynh_print_info "Remove the mysql database"
 
 # Remove a database if it exists, along with the associated user
-ynh_mysql_remove_db $db_user $db_name
+ynh_mysql_remove_db $db_name $db_name
```
maniackcrudelis: Using db_user is just a way to keep this command understandable.
Anyway, at the next release, we should have getopts argument which will give that helper clear named options.

yalh76: OK, fine, I remove the pull request ^^

PR #72, opened 2019-02-17
Adding license informations

yalh76:


JimboJoe:
I'm wondering if it's a good idea to have this information, provided it's already in the manifest... :thinking:

yalh76:
Well the [package_linter](https://github.com/YunoHost/package_linter/blob/master/package_linter.py) ask for....
capture

alexAubin:
Note that it's only if the license field is "free", "non-free" or "dep-non-free" : https://github.com/YunoHost/package_linter/blob/master/package_linter.py#L219-L220

If you put e.g. GPL3.0 (dunno the exact syntax) it's okay

yalh76:
My bad, I remove that Pull Request

alexAubin:
Well maybe we can make the package linter's message that you mention more explicit about this tho. Maybe it's not so obvious that the alternative is to put something else than just "free" in this field

yalh76:
I think the package_linter is enough explicit, I just read it too quickly
`[YEP-1.3] The use of 'free' in license field implies to write something about the license in your README.md`

PR #74, opened 2019-02-26
comment the usage of _common.sh

Mickael-Martin:
For noobs, like me.
I doesn't found on documentation why the source is different on my "normal" scripts (install, remove, upgrade) and my "save" files (backup and restore). This comment add this small, but precious information.


Mickael-Martin:
so merging ?

Mickael-Martin:
thanks \o/

PR #75, opened 2019-02-27
PR to not have misunderstanding between db_user and db_name

toitoinebzh:
I propose these modifications in order 
 * to not have misunderstanding between db_user and db_name 
 * to be consistent with documentation https://yunohost.org/#/packaging_apps_helpers_en

### DIFF
--- a/scripts/backup
+++ b/scripts/backup
@@ -33,7 +33,7 @@ domain=$(ynh_app_setting_get $app domain)
 db_name=$(ynh_app_setting_get $app db_name)
 # it is recommended to have db_user=db_name
 # if you made an other choice, you can assign db_user with `db_user=$(ynh_app_setting_get $app db_user)`
-db_user=db_name 
+db_user=$db_name 
 
 #=================================================
 # STANDARD BACKUP STEPS
--- a/scripts/backup
+++ b/scripts/backup
@@ -31,7 +31,9 @@ app=$YNH_APP_INSTANCE_NAME
 final_path=$(ynh_app_setting_get $app final_path)
 domain=$(ynh_app_setting_get $app domain)
 db_name=$(ynh_app_setting_get $app db_name)
-db_user=$(ynh_app_setting_get $app db_user)
+# it is recommended to have db_user=db_name
+# if you made an other choice, you can assign db_user with `db_user=$(ynh_app_setting_get $app db_user)`
+db_user=db_name 
 
 #=================================================
 # STANDARD BACKUP STEPS
--- a/scripts/backup
+++ b/scripts/backup
@@ -31,6 +31,7 @@ app=$YNH_APP_INSTANCE_NAME
 final_path=$(ynh_app_setting_get $app final_path)
 domain=$(ynh_app_setting_get $app domain)
 db_name=$(ynh_app_setting_get $app db_name)
+db_user=$(ynh_app_setting_get $app db_user)
 
 #=================================================
 # STANDARD BACKUP STEPS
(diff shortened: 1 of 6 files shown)


scripts/install
```diff
@@ -117,8 +117,10 @@ ynh_print_info "Creating a MySQL database..."
 ###		- As well as the section "RESTORE THE MYSQL DATABASE" in the restore script
 
 db_name=$(ynh_sanitize_dbid $app)
+db_user="user_"$db_name
```
maniackcrudelis: $db_user is usually $db_name. Why would you have a different name for the user and the database ?

toitoinebzh:
would you prefer  db_user=$db_name ?
I like the "user_" to make sure that my variable are correctly assign. Right now, when I see db_name in a log somewhere, I do not know if it db_name or db_user ...

maniackcrudelis:
Hum, if we really want to separate db_user and db_name, I'd rather prefer to have `db_user=$db_name`
Personally I use `$db_name` only, since that the same value.

toitoinebzh:
the idea of example_ynh is to help new developer to write script for yunohost, 
I do not have a lot of experience with yunohost but I like to make sure that there is no misunderstanding between variables.
Should we wait for other people's opinion ?

maniackcrudelis:
Go for `db_user=$db_name` if you think that's not obvious as it is.

toitoinebzh:
done

scripts/backup
```diff
@@ -31,6 +31,7 @@ app=$YNH_APP_INSTANCE_NAME
 final_path=$(ynh_app_setting_get $app final_path)
 domain=$(ynh_app_setting_get $app domain)
 db_name=$(ynh_app_setting_get $app db_name)
+db_user=$(ynh_app_setting_get $app db_user)
```
maniackcrudelis: As db_user=db_name, you can avoid storing its value, and simply use `db_user=$db_name`

toitoinebzh: this is true only if the "user" choose to keep db_user=db_name, if the user make an other choice, scripts won't work

maniackcrudelis: As we try to make a standard with this example app.
There's no reason to complicate the process with another variable for db_user.
So using db_user instead of db_name to make the purpose of it clearer, why not.
But using another value for db_user will be up to the packager, and not recommended.
So I would not incite packagers to do otherwise.

toitoinebzh: what do you think of something like : 

``` bash
final_path=$(ynh_app_setting_get $app final_path)
domain=$(ynh_app_setting_get $app domain)
db_name=$(ynh_app_setting_get $app db_name)
# it is recommended to have db_user=db_name
# if you made an other choice, you can assign db_user with `db_user=$(ynh_app_setting_get $app db_user)`
db_user=db_name 
```

maniackcrudelis: As this is the example app, which needs explanations, why not

toitoinebzh:
I update the file in order to always have db_user=db_name

scripts/backup
```diff
@@ -31,6 +31,9 @@ app=$YNH_APP_INSTANCE_NAME
 final_path=$(ynh_app_setting_get $app final_path)
 domain=$(ynh_app_setting_get $app domain)
 db_name=$(ynh_app_setting_get $app db_name)
+# it is recommended to have db_user=db_name
+# if you made an other choice, you can assign db_user with `db_user=$(ynh_app_setting_get $app db_user)`
+db_user=db_name
```
maniackcrudelis: You forgot the '$' before db_name.

toitoinebzh:
I corrected $db_name

yalh76:
Most of this PR has been implemented in https://github.com/YunoHost/example_ynh/pull/86

Maybe we can close this one :)

maniackcrudelis:
Feel free to reopen if needed.

PR #77, opened 2019-03-07
Normalizing url ain't needed anymore

alexAubin:
It's been a while since you actually don't need to normalize the path anymore ... it's done automatically by YunoHost (and also for the change_url case)

https://github.com/YunoHost/yunohost/blob/stretch-testing/src/yunohost/app.py#L2286

You can check it by launching an app script and feeding values like `////trololo////` or `foo/` or `/////` and printing the value of `$path_url` before running the helper...

### DIFF
--- a/scripts/change_url
+++ b/scripts/change_url
@@ -34,15 +34,6 @@ final_path=$(ynh_app_setting_get $app final_path)
 #db_pwd=$(ynh_app_setting_get $app db_pwd)
 
 #=================================================
-# CHECK THE SYNTAX OF THE PATHS
-#=================================================
-
-test -n "$old_path" || old_path="/"
-test -n "$new_path" || new_path="/"
-new_path=$(ynh_normalize_url_path $new_path)
-old_path=$(ynh_normalize_url_path $old_path)
-
-#=================================================
 # CHECK WHICH PARTS SHOULD BE CHANGED
 #=================================================
 
--- a/scripts/install
+++ b/scripts/install
@@ -53,9 +53,6 @@ ynh_print_info "Validating installation parameters..."
 final_path=/var/www/$app
 test ! -e "$final_path" || ynh_die "This path already contains a folder"
 
-# Normalize the url path syntax
-path_url=$(ynh_normalize_url_path $path_url)
-
 # Register (book) web path
 ynh_webpath_register $app $domain $path_url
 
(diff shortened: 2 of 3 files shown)


maniackcrudelis:
I didn't know it was now fixed into the core :open_mouth:
Is it also fixed for a blank value, case for change_url ?

Josue-T:
> Is it also fixed for a blank value, case for change_url ?

Yes

For install it's ok. But for the upgrade, I think it still needed, because the stored settings "path" stay as it is.

alexAubin:
Meh yes I guess you're right about the upgrade script ... though this goes back to July 2017 / 2.7 (c.f. https://github.com/YunoHost/yunohost/pull/304 )

I'll put back the lines in the upgrade script with a note about it

>Is it also fixed for a blank value, case for change_url ?

I'm pretty sure it is, yes, because this line will yield `'/'` even if the original value is empty. Also yes, the function is called for values fed to the change_url script

https://github.com/YunoHost/yunohost/blob/testing/src/yunohost/domain.py#L285

maniackcrudelis:
That's the example app !
Any app built from this package is a new app. So there's not going to be any problem with an older version.

And as Aleks said, if that's from July 2017, I think we can get rid of it.

alexAubin:
>Any app built from this package is a new app.

Uuuuh yes but we also have people in the app team who use this as a template to update older apps sooo, idk :P

kay0u:
Maybe the check `incorrect_path` should be removed from the CI then?

maniackcrudelis:
Would love to !

scripts/change_url
```diff
@@ -33,15 +33,6 @@ final_path=$(ynh_app_setting_get $app final_path)
...
 #db_pwd=$(ynh_app_setting_get $app db_pwd)
 
-#=================================================
-# CHECK THE SYNTAX OF THE PATHS
-#=================================================
-
-test -n "$old_path" || old_path="/"
-test -n "$new_path" || new_path="/"
-new_path=$(ynh_normalize_url_path $new_path)
-old_path=$(ynh_normalize_url_path $old_path)
```
Josue-T: About the old path I think that we also need to keep the check, in case of a bad stored settings

alexAubin: Those are not fetched from the settings.yml from the app, they are fed directly to the app script via bash variables, which have been cleaned using the `_normalize_domain_and_path` python function so those are good :+1:

alexAubin:
ˢᵐᵒˡ ᵇᵘᵐᵖ

PR #78, opened 2019-04-15
[enh] Apply backup core only policy

zamentur:
I suggest to add this in backup script  with "big" backup dir. It allows user to indicate if they want backup this data or not.

Alternatively, we could add this in python core


maniackcrudelis:
Disagree...

PR #82, opened 2019-05-02
Adding ynh_systemd_action

yalh76:


maniackcrudelis:
Just a small thing, I don't think that part should be into "STANDARD" steps.
Even if I'm wondering if we should keep those titles.

yalh76:
Well.. In the standard steps there are already: SETUP SYSTEMD and ADVERTISE SERVICE IN ADMIN PANEL
So for me it's just logical to start the service you have setuped and advertised in the admin panel ^^

But it's as you want, I don't want to overload example_ynh

maniackcrudelis:
I didn't look at it actually...
Ok, keep it there then.

scripts/install
```diff
@@ -297,6 +297,22 @@ yunohost service add $app --log "/var/log/$app/$app.log"
 # if using yunohost version 3.2 or more in the 'manifest.json', a description can be added
 #yunohost service add $app --description "$app daemon for XXX" --log "/var/log/$app/$app.log"
 
+#=================================================
+# START SYSTEMD SERVICE
+#=================================================
+ynh_script_progression --message="Starting a systemd service..." --time --weight=1
```
kay0u: Maybe you should display: "Starting the service $app..." instead? It seems clearer to me.

yalh76: I can, but I tried to be as "generic" as SETUP SYSTEMD

https://github.com/YunoHost/example_ynh/blob/645a1656e92437c6c62d10c3f767da1ab2549eee/scripts/install#L185-L189

So no opinion on that....

Let me know if I add $app or let it like it is

kay0u: I don't know, for some users, what is systemd may not be clear.

But ok, keep it like that :-)

PR #85, opened 2019-05-02
adding repository links

yalh76:


maniackcrudelis:
Is it the repository of the upstream app ? Or the repository of the package ?

yalh76:
It's for the repository of the upstream app.

For packager to let them go to the upstream app to see if there are new releases, issues etc...

For users to let them see if there is some specific issues

maniackcrudelis:
That's a good idea, I've already seen that in some apps, probably yours I guess.

Could you just change something in the sentence that would get a clearer information about what's expected here. Maybe just say that's for the upstream app.
Especially because we do not have so much examples to show for this.

yalh76:
made a few change to precise upstream.

Yes I already use to add it because I need it and sometime I was just bored to go the the upstream app website, took time to find the repository, not finding it and them go to app.src to retrieve the link to the upstream app repository

PR #86, opened 2019-05-02
fix db_user missing

yalh76:
I do my Maniack ^^

### DIFF
--- a/scripts/change_url
+++ b/scripts/change_url
@@ -31,6 +31,7 @@ final_path=$(ynh_app_setting_get --app=$app --key=final_path)
 
 # Add settings here as needed by your application
 #db_name=$(ynh_app_setting_get --app=$app --key=db_name)
+#db_user=$db_name
 #db_pwd=$(ynh_app_setting_get --app=$app --key=db_pwd)
 
 #=================================================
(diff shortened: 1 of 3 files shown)


maniackcrudelis:
I actually do prefer to use dbname instead of dbuser ;)

yalh76:
> 
> 
> I actually do prefer to use dbname instead of dbuser ;)

me too but as `db_user` has been implemented in some scripts but not all. Or we remove `db_user` in all scripts or have to implement db_user in all scripts ^^

But as there are some apps, for example Mastodon, that don't use dthe same db_user and db_name...

https://github.com/YunoHost-Apps/mastodon_ynh/blob/eeec0653ac3e2c38bf6c46c93fdecf28d8dae1ab/scripts/install#L126-L127

Maybe better to implement db_user in all scripts

maniackcrudelis:
I found that using db_user is useless. Since we usually copy db_name to make db_user.
But using db_user could as well give a clear info about what's we really dealing with.

I have no strong opinion on this.

yalh76:
> 
> 
> I found that using db_user is useless. Since we usually copy db_name to make db_user.
> But using db_user could as well give a clear info about what's we really dealing with.
> 
> I have no strong opinion on this.

Me neither, maybe guys from @YunoHost-Apps/apps-group can give their opinion.

- Option 1: remove db_user variable
- Option 2 : implement db_user in all scripts

kay0u:
¯\\\_(ツ)\_/¯

yalh76:
> 
> 
> ¯\_(ツ)_/¯

MDR... it doesn't help ^^

kay0u:
We can keep this one, I'll close #87

Will be merge in 3 days

PR #87, opened 2019-05-02
removing db_user

yalh76:
or https://github.com/YunoHost/example_ynh/pull/86 to implement db_user in all scripts


kay0u:
I close this PR, we keep this one instead: #86

PR #89, opened 2019-05-09
Update check_process.default

kay0u:


check_process.default
```diff
@@ -27,18 +27,8 @@
 		port_already_use=0
```
kay0u: Perhaps we should remove `incorrect_path=1`?

maniackcrudelis: Yep, I remember that we were supposed to.
But, we should find where's the commit that check that.
And add it here as a justification.

kay0u: Got it!

https://github.com/YunoHost/yunohost/pull/304

kay0u:
https://github.com/YunoHost/package_check/pull/56

kay0u:
Bump

PR #90, opened 2019-05-10
 adding additional ynh_script_progression 

yalh76:


alexAubin:
Meh, my 2 cents but I'd just really encourage to use ynh_info / ynh_script_progression with parsimony ... To me the idea was really to stick to ~10ish messages (maybe 15ish for really complex apps) to explain the "big steps" of what's going on. 

But imho having messages about how the script is backuping configuration file checksum or restoring logrotate configuration is only understandable by packagers and leads to the user being all "meh this is a wall of technical messages I don't understand" :confused:

I dunno, maybe I'm just overthinking about a small detail but meh, just keep in mind that those messages are to be designed for our target public and not for packagers ...

Kinda unrelated, but about this thing of "Advertising service in admin panel" .. it's kinda inacurate / misleading : `yunohost service add` is much more general than just 'having it displayed in the adminpanel', it's a more general thing that the service is known by yunohost so that you can use yunohost *also from the CLI* to interact with the service (start/stop, but also get the logs)

yalh76:
^^ I over reacted, thinking as a packager with debug mode activated :)
I close that PR

alexAubin:
I mean some of them might be good and relevant to keep though, idk, it's just the example app anyway .. that was more a general discussion about how I see the use of ynh_info / progression, maybe other people have different views ...

maniackcrudelis:
In a different way I agree with Aleks.
`ynh_script_progression` is especially useful to show the progression of the script, and avoid being stuck behind the script, wondering if the whole thing is working or not.
So I find it interesting for steps needed some time to be performed.
Here a lot of them would be really quick, adding an info would probably just print a lot of lines a the same time. Not really that interesting.
Also, I agree with Aleks about the idea of giving non-too-technical info about what's going on.

yalh76:
Well, I just implemented ynh_script_progression on mastodon_ynh

1. implementing it took a lot of time. When you package is ready, you have to launch each script to have the time of each steps
1. for mastodon weight is irrelevant 
example with the install script:
* Validating installation parameters..." --weight=2
* Storing installation settings..." --weight=2
* Configuring firewall..." --weight=1
* Installing dependencies..." --weight=86
* Creating a PostgreSQL database..." --weight=5
* Setting up source files..." --weight=5
* Configuring nginx web server..." --weight=3
* Configuring system user..." --weight=4
* Installing Ruby..." --weight=424
* Modifying a config file..." --weight=2
* Installing Mastodon..." --weight=2230
* Setuping a cron job for removing cache..." --weight=1
* Configuring a systemd service..." --weight=5
* Storing the config file checksum..." --weight=1
* Securing files and directories..." --weight=9
* Advertising service in admin panel..." --weight=3
* Starting a systemd service..." --weight=47
* Configuring SSOwat..." --weight=2
* Reloading nginx web server..." --weight=2
* Sending a readme for the admin..." --weight=17

The `Installing Mastodon..." --weight=2230` took 78% of the progress bar...

I'm thinking that this feature will reduce packaging efforts, but just my opinion.

Maybe something more automated, like when doing a package_check with an argument like `--time`, in addition of the log file, will produce a "timer" file that ynh_script_progression will use to automatically retrieve the weights for each steps of each script...

maniackcrudelis:
2230!? Really!
For that kind of situation I reduce the size to not have too big steps.
Like reducing to 120 for this one and 60 for other really big steps and so on.

Also, you may have seen that I'm doing that on all official apps (work still ongoing though). To not try each script one by one, I'm just running package check like usual. That gives me the times for each steps, then I apply while package check is still running.
Usually, at the end I ran it a last time to be sure I didn't broke anything.

But that way that doesn't takes that much time to do.
You could even run it on the dev CI and just read the times afterward.

yalh76:
yes: the install step, that do the compilation of all ruby and javascript files took 00h37m,09s

I also thinking to use package check to retrieve time for the other packages I work for. 

For backup and restore the timer is even more irrelevant because: 

- during backup, a lot of time is took at the end of the script to `Now creating a backup archive from the files collected…`
- during restore, a lot of time is took at the beginning of the script to `Preparing archive for restoration…`

But ok, I will try to use it. That was just to inform you guys that using the ynh_script_progression is quite boring but if it increase users experience, it's ok to try to use it :)

maniackcrudelis:
> 00h37m,09s

Isn't there a way to reduce that time !? That's really huge and uncomfortable for the user.

> For backup and restore the timer is even more irrelevant because:

Indeed, we know that for backup and restore the script is just a small part of the work. But also, the final user won't see the time themselves, only the progress bar.
That's just an indication, so the user knows where he is in the progression of the script.

yalh76:
> Isn't there a way to reduce that time !? That's really huge and uncomfortable for the user.

I think the only way would be to provide an already built version...

maniackcrudelis:
That's all the ruby mess that takes so long I guess, I'm not even sure you could have a built version of that thing.

What an idea anyway to build an app with Ruby **AND** Node.js !? There's probably nothing worse than those 2 things !!!

PR #92, opened 2019-05-17
Php fpm

yalh76:
I propose to change the php-fpm step order.

I'm considering package using php7.1, 7.2, etc...  and installing php7.1, 7.2, etc... during the `INSTALL DEPENDENCIES` step

- during removal it's better to have php-fpm removal before dependencies removal
- during restore we must first install dependencies and then restore php-fpm

### DIFF
--- a/scripts/remove
+++ b/scripts/remove
@@ -52,6 +52,14 @@ ynh_script_progression --message="Removing the MySQL database..." --time --weigh
 ynh_mysql_remove_db --db_user=$db_user --db_name=$db_name
 
 #=================================================
+# REMOVE PHP-FPM CONFIGURATION
+#=================================================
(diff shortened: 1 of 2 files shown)


maniackcrudelis:
If you use php 7.1 or more, you're suppose to use the specific helper dedicated to.
Your dependencies should then be installed or removed by this helper, not `ynh_install_app_dependencies`

yalh76:
> 
> 
> If you use php 7.1 or more, you're suppose to use the specific helper dedicated to.
> Your dependencies should then be installed or removed by this helper, not `ynh_install_app_dependencies`

Yes but I usually put `ynh_install_php --phpversion` at `DEPENDENCIES ` Step

maniackcrudelis:
Would be better to use a specific part for this step.

yalh76:
OK, I will do that

PR #94, opened 2019-06-02
Add link to packaging helpers

decentral1se:
As recommended by @kay0u.

Rendered version:

> https://github.com/YunoHost/example_ynh/blob/3046ec3fdc9ddcbf198cdae8d08c7f2980c32cc4/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@
 - Edit `conf/nginx.conf` file to match application prerequisites.
 - Edit `manifest.json` with application specific information.
 - Edit the `install`, `upgrade`, `remove`, `backup`, and `restore` scripts.
+  - Using the [script helpers documentation.](https://helpers.yunohost.org/)
 - Add a `LICENSE` file for the package.
 - Edit `README.md` and README_fr.md.
 


README.md
```diff
@@ -3,6 +3,7 @@
 - Edit `conf/nginx.conf` file to match application prerequisites.
 - Edit `manifest.json` with application specific information.
 - Edit the `install`, `upgrade`, `remove`, `backup`, and `restore` scripts.
+  - Using the [script helpers documentation](https://yunohost.org/#/packaging_apps_helpers)
```
JimboJoe: On my French browser, the page doesn't exist... whereas http://helpers.yunohost.org/ points to the English page (no French page yet). Shall we use that link instead (is it official)?

kay0u: That's weird... But if it's working

decentral1se: Uhhhh, should I use this other new URL?

decentral1se: OK, fixed in https://github.com/YunoHost/example_ynh/pull/94/commits/4327e037eec95a0732cc94edd370985d934747b3. Points to the helpers ... URL.

decentral1se:
OK, this should be good to go then.

PR #96, opened 2019-06-19
Backup before changing url.

maniackcrudelis:
## Problem
- *If a changeurl script fails, the app will end in a none fully moved state. Forcing the user to fix it by himself.*

## Solution
- *Use `ynh_backup_before_upgrade` into the changeurl script.*
- *Remove the nginx config file for the new domain, the remove script will try to remove the old one, as the modification of the config would have happened at the end.*

## PR Status
- [x] Code finished.
- [x] Can be reviewed and tested.

### DIFF
--- a/scripts/change_url
+++ b/scripts/change_url
@@ -35,6 +35,23 @@ final_path=$(ynh_app_setting_get --app=$app --key=final_path)
 #db_pwd=$(ynh_app_setting_get --app=$app --key=db_pwd)
 
 #=================================================
+# BACKUP BEFORE UPGRADE THEN ACTIVE TRAP
+#=================================================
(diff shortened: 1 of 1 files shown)


kay0u:
What do you think about change the name of `ynh_backup_before_upgrade` in `ynh_backup_before`? It could be used with actions and config that will become official one day \o/

(We can add an option --action=upgrade to separate backups if needed)

And change `ynh_restore_upgradebackup` to `ynh_restore_backup` accordingly

maniackcrudelis:
That would mean change it in every scripts as well, is it really worth it?

kay0u:
I don't know, maybe not. The naming triggered me

PR #98, opened 2019-06-28
Adding double quotes around variables

liberodark:
## Problem
Have not good synthaxe is not very easy to make a 100% proper bash
The idea is make a more better bash script for help packager to make better bash

## Solution

## PR Status
- [x] Code finished.

## Package_check results
---
*If you have access to [App Continuous Integration for packagers](https://yunohost.org/#/packaging_apps_ci) you can provide a link to the package_check results like below, replacing '-NUM-' in this link by the PR number and USERNAME by your username on the ci-apps-dev. Or you provide a screenshot or a pastebin of the results*

[Build Status/badge/icon)](https://ci-apps-dev.yunohost.org/jenkins/job/APP_ynh%20PR-NUM-%20(USERNAME)/)

### DIFF
--- a/scripts/backup
+++ b/scripts/backup
@@ -28,9 +28,9 @@ ynh_script_progression --message="Loading installation settings..." --time --wei
 
 app=$YNH_APP_INSTANCE_NAME
 
-final_path=$(ynh_app_setting_get --app=$app --key=final_path)
-domain=$(ynh_app_setting_get --app=$app --key=domain)
-db_name=$(ynh_app_setting_get --app=$app --key=db_name)
+final_path=$(ynh_app_setting_get --app="$app" --key=final_path)
+domain=$(ynh_app_setting_get --app="$app" --key=domain)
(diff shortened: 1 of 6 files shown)


maniackcrudelis:
You don't need that many double quotes everywhere !
Double quotes should be used when needed, not everywhere to be sure it would works. Especially since it won't !

Double quotes have to be used when
- You expect a variable with special characters or spaces.
- You use `test -n` or `test -z`
- You use a string comparison.

Otherwise, double quotes shouldn't be used, it may raise errors, especially if you use them for digit comparisons and can failed cli command if not used correctly.
Putting double quotes everywhere is not a good solution, you have to know and understand where and why use them.

liberodark:
Now this bash is not clean without double quote and no have make double quote everywhere.
An so I also know how to use it but it's much cleaner for a more compatible code.
Make tests you'll see that I'm right.

kay0u:
Closing for non activity

PR #99, opened 2019-07-11
Remove reference to Jessie CI ? :|

alexAubin:
I mean ... zbleh ... buster got released ...


maniackcrudelis:
Yep but we don't have buster support for now ;)

alexAubin:
Come on folks ... at which point did we decided that we were two supports at least two debian version ... Jessie is dead since 2 years now, Yunohost migrated away from it more than 1 year ago ... how many users do we still have on Jessie ... and how many apps are still compatible ...

PR #102, opened 2019-10-02
ShellCheck of the scripts

moutonjr:
Using shellcheck to proofcheck the sanity of the scripts.

Note that there are two sourcing failed on example scripts (backup and restore). Rest of changes are harmless.

Best would be to apply shellcheck for a new YEP ?

See https://forum.yunohost.org/c/contribute-room/apps-packaging/ topic about this.

[Build Status/badge/icon)](https://ci-apps-dev.yunohost.org/jenkins/job/APP_ynh%20PR-NUM-%20(USERNAME)/)

### DIFF
--- a/scripts/_common.sh
+++ b/scripts/_common.sh
@@ -5,7 +5,7 @@
 #=================================================
 
 # dependencies used by the app
-pkg_dependencies="deb1 deb2"
+export pkg_dependencies="deb1 deb2"
 
 #=================================================
 # PERSONAL HELPERS
--- a/scripts/backup
+++ b/scripts/backup
@@ -7,7 +7,7 @@
 #=================================================
 
 #Keep this path for calling _common.sh inside the execution's context of backup and restore scripts
-source ../settings/scripts/_common.sh
+source _common.sh
 source /usr/share/yunohost/helpers
 
 #=================================================
@@ -28,9 +28,9 @@ ynh_script_progression --message="Loading installation settings..." --time --wei
 
 app=$YNH_APP_INSTANCE_NAME
 
-final_path=$(ynh_app_setting_get --app=$app --key=final_path)
-domain=$(ynh_app_setting_get --app=$app --key=domain)
-db_name=$(ynh_app_setting_get --app=$app --key=db_name)
+final_path=$(ynh_app_setting_get --app="$app" --key=final_path)
+domain=$(ynh_app_setting_get --app="$app" --key=domain)
+db_name=$(ynh_app_setting_get --app="$app" --key=db_name)
 
 #=================================================
 # STANDARD BACKUP STEPS
@@ -39,7 +39,7 @@ db_name=$(ynh_app_setting_get --app=$app --key=db_name)
 #=================================================
 ynh_script_progression --message="Stopping a systemd service..." --time --weight=1
 
-ynh_systemd_action --service_name=$app --action="stop" --log_path="/var/log/$app/$app.log"
+ynh_systemd_action --service_name="$app" --action="stop" --log_path="/var/log/$app/$app.log"
 
 #=================================================
 # BACKUP THE APP MAIN DIR
@@ -104,7 +104,7 @@ ynh_backup --src_path="/etc/cron.d/$app"
 #=================================================
 ynh_script_progression --message="Starting a systemd service..." --time --weight=1
 
-ynh_systemd_action --service_name=$app --action="start" --log_path="/var/log/$app/$app.log"
+ynh_systemd_action --service_name="$app" --action="start" --log_path="/var/log/$app/$app.log"
 
 #=================================================
 # END OF SCRIPT
--- a/scripts/change_url
+++ b/scripts/change_url
@@ -27,7 +27,8 @@ app=$YNH_APP_INSTANCE_NAME
 ynh_script_progression --message="Loading installation settings..." --time --weight=1
 
 # Needed for helper "ynh_add_nginx_config"
-final_path=$(ynh_app_setting_get --app=$app --key=final_path)
+final_path=$(ynh_app_setting_get --app="$app" --key=final_path)
+export final_path
 
 # Add settings here as needed by your application
 #db_name=$(ynh_app_setting_get --app=$app --key=db_name)
@@ -57,7 +58,7 @@ fi
 #=================================================
 ynh_script_progression --message="Stopping a systemd service..." --time --weight=1
 
-ynh_systemd_action --service_name=$app --action="stop" --log_path="/var/log/$app/$app.log"
+ynh_systemd_action --service_name="$app" --action="stop" --log_path="/var/log/$app/$app.log"
 
 #=================================================
 # MODIFY URL IN NGINX CONF
@@ -72,8 +73,8 @@ then
 	# Make a backup of the original nginx config file if modified
 	ynh_backup_if_checksum_is_different --file="$nginx_conf_path"
 	# Set global variables for nginx helper
-	domain="$old_domain"
-	path_url="$new_path"
+	export domain="$old_domain"
+	export path_url="$new_path"
 	# Create a dedicated nginx config
 	ynh_add_nginx_config
 fi
@@ -83,7 +84,7 @@ if [ $change_domain -eq 1 ]
 then
 	# Delete file checksum for the old conf file location
 	ynh_delete_file_checksum --file="$nginx_conf_path"
-	mv $nginx_conf_path /etc/nginx/conf.d/$new_domain.d/$app.conf
+	mv "$nginx_conf_path" "/etc/nginx/conf.d/$new_domain.d/$app.conf"
 	# Store file checksum for the new config file location
 	ynh_store_file_checksum --file="/etc/nginx/conf.d/$new_domain.d/$app.conf"
 fi
@@ -101,7 +102,7 @@ fi
 #=================================================
 ynh_script_progression --message="Starting a systemd service..." --time --weight=1
 
-ynh_systemd_action --service_name=$app --action="start" --log_path="/var/log/$app/$app.log"
+ynh_systemd_action --service_name="$app" --action="start" --log_path="/var/log/$app/$app.log"
 
 #=================================================
 # RELOAD NGINX
--- a/scripts/install
+++ b/scripts/install
@@ -29,7 +29,7 @@ path_url=$YNH_APP_ARG_PATH
 admin=$YNH_APP_ARG_ADMIN
 is_public=$YNH_APP_ARG_IS_PUBLIC
 language=$YNH_APP_ARG_LANGUAGE
-password=$YNH_APP_ARG_PASSWORD
+#password=$YNH_APP_ARG_PASSWORD # If useful later ?
 
 ### If it's a multi-instance app, meaning it can be installed several times independently
 ### The id of the app as stated in the manifest is available as $YNH_APP_ID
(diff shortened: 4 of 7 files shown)


scripts/_common.sh
```diff
@@ -5,7 +5,7 @@
 #=================================================
 
 # dependencies used by the app
-pkg_dependencies="deb1 deb2"
+export pkg_dependencies="deb1 deb2"
```
maniackcrudelis: `export` make the variable available into environment level.
We do not use them for that purpose, here the variable is available at the entire script level, which is enough as this script will be sourced by the other scripts.

scripts/backup
```diff
@@ -7,7 +7,7 @@
 #=================================================
 
 #Keep this path for calling _common.sh inside the execution's context of backup and restore scripts
-source ../settings/scripts/_common.sh
+source _common.sh
```
maniackcrudelis: Into the scope of the script, when it's running, the path was correct the way it was.

scripts/backup
```diff
@@ -28,9 +28,9 @@ ynh_script_progression --message="Loading installation settings..." --time --wei
 
 app=$YNH_APP_INSTANCE_NAME
 
-final_path=$(ynh_app_setting_get --app=$app --key=final_path)
-domain=$(ynh_app_setting_get --app=$app --key=domain)
-db_name=$(ynh_app_setting_get --app=$app --key=db_name)
+final_path=$(ynh_app_setting_get --app="$app" --key=final_path)
+domain=$(ynh_app_setting_get --app="$app" --key=domain)
+db_name=$(ynh_app_setting_get --app="$app" --key=db_name)
```
maniackcrudelis: `app` is `YNH_APP_INSTANCE_NAME` which is the `id` of the app with a incremental number attach to it.
It has to follow that regex: `^[a-z1-9]((_|-)?[A-z1-9])+$`
https://github.com/YunoHost/doc/blob/master/packaging_apps_guidelines.md#yep-11

Using double quotes here is therefor useless.
Using them would be as well harmless. But not knowing when and why you should use them is harmful.
So I rather prefer not to use double quotes when not needed instead of putting them everywhere without knowing why we should or should not use them.

scripts/change_url
```diff
@@ -27,7 +27,8 @@ app=$YNH_APP_INSTANCE_NAME
 ynh_script_progression --message="Loading installation settings..." --time --weight=1
 
 # Needed for helper "ynh_add_nginx_config"
-final_path=$(ynh_app_setting_get --app=$app --key=final_path)
+final_path=$(ynh_app_setting_get --app="$app" --key=final_path)
+export final_path
```
maniackcrudelis: There's no purpose here to export that variable, which is not going to be use anywhere into the scope of that script.

scripts/change_url
```diff
@@ -72,8 +73,8 @@ then
 	# Make a backup of the original nginx config file if modified
 	ynh_backup_if_checksum_is_different --file="$nginx_conf_path"
 	# Set global variables for nginx helper
-	domain="$old_domain"
-	path_url="$new_path"
+	export domain="$old_domain"
+	export path_url="$new_path"
```
maniackcrudelis: Same here, no purpose for a export.

scripts/change_url
```diff
@@ -83,7 +84,7 @@ if [ $change_domain -eq 1 ]
 then
 	# Delete file checksum for the old conf file location
 	ynh_delete_file_checksum --file="$nginx_conf_path"
-	mv $nginx_conf_path /etc/nginx/conf.d/$new_domain.d/$app.conf
+	mv "$nginx_conf_path" "/etc/nginx/conf.d/$new_domain.d/$app.conf"
```
maniackcrudelis: As for `app`, there can't be here any characters that would break bash.

scripts/install
```diff
@@ -29,7 +29,7 @@ path_url=$YNH_APP_ARG_PATH
 admin=$YNH_APP_ARG_ADMIN
 is_public=$YNH_APP_ARG_IS_PUBLIC
 language=$YNH_APP_ARG_LANGUAGE
-password=$YNH_APP_ARG_PASSWORD
+#password=$YNH_APP_ARG_PASSWORD # If useful later ?
```
maniackcrudelis: The purpose of this repository is to show how an app package work. Not to be fully working as it is.
The manifest has a [password field](https://github.com/YunoHost/example_ynh/blob/master/manifest.json#L75-L87), therefor we should have this line to get it from the manifest. Whether it's used or not later in the script is not a problem.

scripts/install
```diff
@@ -62,18 +62,18 @@ final_path=/var/www/$app
...
-ynh_app_setting_set --app=$app --key=domain --value=$domain
-ynh_app_setting_set --app=$app --key=path --value=$path_url
-ynh_app_setting_set --app=$app --key=admin --value=$admin
-ynh_app_setting_set --app=$app --key=is_public --value=$is_public
-ynh_app_setting_set --app=$app --key=language --value=$language
+ynh_app_setting_set --app="$app" --key=domain --value="$domain"
+ynh_app_setting_set --app="$app" --key=path --value="$path_url"
+ynh_app_setting_set --app="$app" --key=admin --value="$admin"
+ynh_app_setting_set --app="$app" --key=is_public --value="$is_public"
+ynh_app_setting_set --app="$app" --key=language --value="$language"
```
maniackcrudelis: Same again here, before and after, all those variables don't have double quotes because they don't need any.
If, because it could happen, a variable may really need some somewhere, I would be glad to add them.

scripts/install
```diff
@@ -106,7 +106,7 @@ ynh_script_progression --message="Installing dependencies..." --time --weight=1
 ###		- As well as the section "REINSTALL DEPENDENCIES" in the restore script
 ###		- And the section "UPGRADE DEPENDENCIES" in the upgrade script
 
-ynh_install_app_dependencies $pkg_dependencies
+ynh_install_app_dependencies "$pkg_dependencies"
```
maniackcrudelis: And here, furthermore, that's an error. Again a packager should understand why use them or not.
`ynh_install_app_dependencies` get its arguments from the command line with `$@`, see [the helper](https://github.com/YunoHost/yunohost/blob/stretch-unstable/data/helpers.d/apt#L207), using double quote here would send only one argument to the helper where it expects one or many.
Here it would be safe because the helper gets its argument and store them into an unique variable. But it could as well be an array, broken by such practice.

scripts/install
```diff
@@ -228,9 +228,9 @@ ynh_script_progression --message="Finalizing installation..." --time --weight=1
 ynh_local_curl "/INSTALL_PATH" "key1=value1" "key2=value2" "key3=value3"
 
 # Remove the public access
-if [ $is_public -eq 0 ]
+if [ "$is_public" -eq 0 ]
```
maniackcrudelis: `$is_public` is not `"$is_public"`, the first one is an int, the second one a string.
You may have an error from bash with such a condition.

scripts/remove
```diff
@@ -16,11 +16,11 @@ ynh_script_progression --message="Loading installation settings..." --time --wei
 
 app=$YNH_APP_INSTANCE_NAME
 
-domain=$(ynh_app_setting_get --app=$app --key=domain)
-port=$(ynh_app_setting_get --app=$app --key=port)
-db_name=$(ynh_app_setting_get --app=$app --key=db_name)
+#domain=$(ynh_app_setting_get --app="$app" --key=domain) If useful?
```
maniackcrudelis: `ynh_remove_nginx_config`, later in the script use it.

scripts/install
moutonjr: hmm. not for this one:
```
vagrant@yunohost:~$ [ $is_public -eq 0 ] && echo TRUE
TRUE
vagrant@yunohost:~$ [ "$is_public" -eq 0 ] && echo TRUE
TRUE
vagrant@yunohost:~$ [ "$is_public" -eq "0" ] && echo TRUE
TRUE
```

maniackcrudelis: Indeed, not for THIS one

scripts/install
moutonjr: Not really if `$@` is used properly:
```
vagrant@yunohost:~$ function my_installer() { for i in $@; do echo "Install " $i; done }
vagrant@yunohost:~$ my_installer "packet1 packet2 packet3"
Install  packet1
Install  packet2
Install  packet3
vagrant@yunohost:~$ myvar="packet1 packet2 packet3"
vagrant@yunohost:~$ my_installer "$myvar"
Install  packet1
Install  packet2
Install  packet3
```

maniackcrudelis: Because `for` will split the variable by its spaces, as `while` would do. Still, that's one argument instead of many as it's expected.

scripts/install
moutonjr: you are right, that was what [Aleks ](https://github.com/alexAubin)told me in the  [related topic](https://forum.yunohost.org/t/best-pracices-using-shellcheck-to-sanitize-scripts/9152/3). This is just best practice, not useful, but an example to make the repo compatible with bash standards.

Proof-checking such vars are useless, event such scripts but it's just in that context. applying rule for every variable is a good practice anyway.

scripts/change_url
moutonjr: I was then confused by the upper comment : "` # Needed for helper "ynh_add_nginx_config" `" which is indeed outside of the script.

scripts/backup
moutonjr: My bad, I was confused by the runtime environment of this, which is different then the other scripts.

scripts/install
maniackcrudelis: Applying blindly good practices doesn't sound like a good idea to me.

scripts/backup
maniackcrudelis: Indeed, and it's not really obvious than those 2 scripts are running into a different environment.

scripts/_common.sh
moutonjr: The purpose of `export` is to indicate that this variable will be used outside of the script. Per se, it doesn't do anything more than plain vars if the bash file is sourced like here.

Yet the var should be used across script files, indeed the _helper_ of installation.

But I don't agree with your analysis of the behavior, the environment level is not affected by export if not sourced:
```
vagrant@yunohost:~$ echo "export NAME=johndoe" > sourced_bash.sh
vagrant@yunohost:~$ echo 'source sourced_bash.sh' > script.sh
vagrant@yunohost:~$ echo 'echo "$NAME"' >> script.sh
vagrant@yunohost:~$ bash script.sh
johndoe
vagrant@yunohost:~$ echo $NAME
```


In short: still useless but good practice (https://github.com/koalaman/shellcheck/wiki/SC2034)

scripts/install
moutonjr: @maniackcrudelis I got your point of your review, don't worry.

scripts/install
moutonjr: "Blindly"... no need to get that harsh.

moutonjr:
OK let's cut this, bash lintering is not in the topics of this repository.

scripts/install
maniackcrudelis: No offense
I'm not saying that for you, but more for all packagers that would went after. I rather prefer them to learn from their mistake instead of following rules without understanding them.
Because if they don't know, when the issue will come (and it always...), that will be worse.

moutonjr: I don't know. When is the bad practice harmful, and when is it not, especially for non-fluent bash users? That's the purpose of linting the code and using tools like ShellCheck to explain the problematic with details and propose solution, even if not risk-relevant in some cases. It helps build habits?

PR #110, opened 2020-02-08
Set the default php version to use

maniackcrudelis:
## Problem
- *php7.0 is the default version only for stretch. Buster use 7.3 and bullseye will probably use another one*

## Solution
- *Set the version of php used by the app, so it can be easily modified without having to change each apps.*
- ~*Move nginx after php during the upgrade, so the nginx config will be set according to php in case of migration to an upper version.*~

**This PR should wait for https://github.com/YunoHost/yunohost/pull/880 before being merged.**

## PR Status
- [ ] Code finished.

### DIFF
--- a/scripts/_common.sh
+++ b/scripts/_common.sh
@@ -5,7 +5,7 @@
 #=================================================
 
 # dependencies used by the app
-pkg_dependencies="deb1 deb2"
+pkg_dependencies="deb1 deb2 php$YNH_DEFAULT_PHP_VERSION-deb1 php$YNH_DEFAULT_PHP_VERSION-deb2"
 
 #=================================================
 # PERSONAL HELPERS
--- a/scripts/backup
+++ b/scripts/backup
@@ -31,6 +31,7 @@ app=$YNH_APP_INSTANCE_NAME
 final_path=$(ynh_app_setting_get --app=$app --key=final_path)
 domain=$(ynh_app_setting_get --app=$app --key=domain)
 db_name=$(ynh_app_setting_get --app=$app --key=db_name)
+phpversion=$(ynh_app_setting_get --app=$app --key=phpversion)
 
 #=================================================
 # STANDARD BACKUP STEPS
@@ -60,7 +61,7 @@ ynh_backup --src_path="/etc/nginx/conf.d/$domain.d/$app.conf"
 #=================================================
 ynh_script_progression --message="Backing up php-fpm configuration..." --time --weight=1
 
-ynh_backup --src_path="/etc/php/7.0/fpm/pool.d/$app.conf"
+ynh_backup --src_path="/etc/php/$phpversion/fpm/pool.d/$app.conf"
 
 #=================================================
 # BACKUP THE MYSQL DATABASE
--- a/scripts/install
+++ b/scripts/install
@@ -75,6 +75,8 @@ ynh_app_setting_set --app=$app --key=admin --value=$admin
 ynh_app_setting_set --app=$app --key=is_public --value=$is_public
 ynh_app_setting_set --app=$app --key=language --value=$language
 
+ynh_app_setting_set --app=$app --key=phpversion --value=$YNH_DEFAULT_PHP_VERSION
+
 #=================================================
 # STANDARD MODIFICATIONS
 #=================================================
@@ -153,7 +155,7 @@ ynh_script_progression --message="Configuring nginx web server..." --time --weig
 ### `ynh_add_nginx_config` will use the file conf/nginx.conf
 
 # Create a dedicated nginx config
-ynh_add_nginx_config
+ynh_add_nginx_config php_version
 
 #=================================================
 # CREATE DEDICATED USER
--- a/scripts/install
+++ b/scripts/install
@@ -75,8 +75,6 @@ ynh_app_setting_set --app=$app --key=admin --value=$admin
 ynh_app_setting_set --app=$app --key=is_public --value=$is_public
 ynh_app_setting_set --app=$app --key=language --value=$language
 
-ynh_app_setting_set --app=$app --key=phpversion --value=$YNH_DEFAULT_PHP_VERSION
-
 #=================================================
 # STANDARD MODIFICATIONS
 #=================================================
@@ -155,7 +153,7 @@ ynh_script_progression --message="Configuring nginx web server..." --time --weig
 ### `ynh_add_nginx_config` will use the file conf/nginx.conf
 
 # Create a dedicated nginx config
-ynh_add_nginx_config php_version
+ynh_add_nginx_config
 
 #=================================================
 # CREATE DEDICATED USER
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -18,7 +18,7 @@ location __PATH__/ {
   try_files $uri $uri/ index.php;
   location ~ [^/]\.php(/|$) {
     fastcgi_split_path_info ^(.+?\.php)(/.*)$;
-    fastcgi_pass unix:/var/run/php/php7.0-fpm-__NAME__.sock;
+    fastcgi_pass unix:/var/run/php/php__PHPVERSION__-fpm-__NAME__.sock;
 
     fastcgi_index index.php;
     include fastcgi_params;
--- a/conf/php-fpm.conf
+++ b/conf/php-fpm.conf
@@ -33,7 +33,7 @@ group = __USER__
 ;                            (IPv6 and IPv4-mapped) on a specific port;
 ;   '/path/to/unix/socket' - to listen on a unix socket.
 ; Note: This value is mandatory.
-listen = /var/run/php/php7.0-fpm-__NAMETOCHANGE__.sock
+listen = /var/run/php/php__PHPVERSION__-fpm-__NAMETOCHANGE__.sock
 
 ; Set listen(2) backlog.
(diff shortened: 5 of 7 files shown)


maniackcrudelis:
Still an issue in that code, in case of a migration... $phpversion won't be updated before passed to the nginx helper...
Which lead to another question, should we set $phpversion as $PHPVERSION ?

scripts/install
```diff
@@ -153,7 +155,7 @@ ynh_script_progression --message="Configuring nginx web server..." --time --weig
 ### `ynh_add_nginx_config` will use the file conf/nginx.conf
 
 # Create a dedicated nginx config
-ynh_add_nginx_config
+ynh_add_nginx_config php_version
```
kay0u: The variable `$php_version` is not defined here, we should have somewhere above

```
php_version=$YNH_DEFAULT_PHP_VERSION
```

https://github.com/YunoHost/yunohost/blob/62a98eb0733bbc4e396a071cf07d6a38fc2cd1a3/data/helpers.d/nginx#L52

scripts/install
```diff
@@ -75,6 +75,8 @@ ynh_app_setting_set --app=$app --key=admin --value=$admin
 ynh_app_setting_set --app=$app --key=is_public --value=$is_public
 ynh_app_setting_set --app=$app --key=language --value=$language
 
+ynh_app_setting_set --app=$app --key=phpversion --value=$YNH_DEFAULT_PHP_VERSION
```
kay0u: For now `$YNH_DEFAULT_PHP_VERSION` is not defined right?

I can add it if needed

kay0u:
> Still an issue in that code, in case of a migration... $phpversion won't be updated before passed to the nginx helper...
> Which lead to another question, should we set $phpversion as $PHPVERSION ?

I'm sorry I don't get the problem O.o

scripts/install
kay0u: I didn't see this work:

https://github.com/YunoHost/yunohost/blob/4996c49cf641095aa986dc6ee63daf4645a63618/data/helpers.d/php#L3-L8

We can do it there I suppose, but IMO it would be easier if all the env var came from the same place (the core).

scripts/install
maniackcrudelis: Indeed, the variable is added to the config, but not properly initialized...
Same as the question I asked before, shouldn't we run the php helper before and set the variable as global by the way ?

scripts/install
maniackcrudelis: The idea indeed is to set it from the php helper.
But so far, no it isn't defined, which wouldn't work with the current code.
However, this [PR](https://github.com/YunoHost/yunohost/pull/879), could as well be merged into stretch.

Not sure what you mean about the variable into the core.

maniackcrudelis:
That's related to https://github.com/YunoHost/yunohost/pull/879, that the php helper that will change the version of php to use. So if we run the nginx helper before, it will use the previous version of php, and then the php helper will set another one, ending with nginx not able to find the socket.

scripts/install
kay0u: I mean, at the same place as `YNH_APP_ID` `YNH_APP_INSTANCE_NAME`...

maniackcrudelis: As you prefer, anyway, we're not going to change it often. Feel free to modify the PR accordingly.

kay0u:
Doesn't this problem only appear if we reload the nignx service before running php helper?

maniackcrudelis:
Except that nginx should use $YNH_DEFAULT_PHP_VERSION unless php do the same. That's why we have to know which version php use before set the version for nginx.

kay0u:
IMO, we should use `$YNH_DEFAULT_PHP_VERSION` if no other version is specified in both case. `$YNH_DEFAULT_PHP_VERSION` is defined at the beginning of the script (or is an env var).

kay0u:
Hmm, perhaps we could create a var `php_used_by_fpm` when we run PHP helper, and use it inside nginx helper? (I think it's what you've been trying to say all along, but I have a slow brain ~sometimes~)

maniackcrudelis:
I agree with you (I fixed my previous post by the way !)
We should use $YNH_DEFAULT_PHP_VERSION is no specific version is specified. But here my concern is, first that in case of migration, the default version may be changed by the php helper, and second that the app may not use the default version.
It appears to me that it's better to say once the version the app will use. And so, the php helper should defined that. Otherwise, that would mean to set it for all apps.

So my idea is either we have a simple app that use the default version. The version is defined by $YNH_DEFAULT_PHP_VERSION.
But a migration to a new version will modify this default version.
Or we have an app that use a specific version of php. In that case, the version will be set using the variable $phpversion and pass to the php helper.
But the app should do the same for nginx.

Here the issue I see.
In the first case, we use $YNH_DEFAULT_PHP_VERSION  but $phpversion has never been set since it isn't used. Only the php helper will do it, but as a local variable only.
So here, the nginx helper should set either $phpversion or $YNH_DEFAULT_PHP_VERSION as default.

In the second case, $phpversion is set and will be passed to the php helper.
And as well to the nginx helper.

[...]
Writing all of this I realize that it doesn't really matters...
On one hand, we do not set any specific version, so we use $YNH_DEFAULT_PHP_VERSION, on the other, we set a specific version, which will be used by php and can be used as well for nginx. We should just add a line into the nginx helper to take care of this field like any other.
The only thing is to put the nginx part after the php part into the upgrade script to handle properly the case of migration. That's all...

maniackcrudelis:
My brain seems to be slow as well today...
If we use $YNH_DEFAULT_PHP_VERSION, it doesn't matter which one of nginx or php run first... Both will use that value, and so... why use $phpversion... ?

You know what, I'm going to have a break, do my workout and be back with maybe my brain clearer.

maniackcrudelis:
So...

Let's have things clear:
- We do need phpversion into the config of each app, whether it's the default version or not.
It will be used for remove, restore and backup scripts that have to know which version is used, even if not the **current** default version.
- At the same time, we do not need to set it during the install script, the php helper should do it for us. (But it doesn't for now...)
- In case of a default version, the nginx script should still know which version it is. I think set a env variable into the php helper is a good idea.
It could even be better to set it at the beginning of the php file, so it will be available at any time into the install and upgrade scripts (especially for nginx)
- In case of a non default version, the packager could just modify the env variable.
[...] Except that we load the _common.sh **before** the official helpers... Shit...

kay0u:
> In case of a non default version, the packager could just modify the env variable.
[...] Except that we load the _common.sh before the official helpers... Shit...

Can we imagine to modify the env variable here?

https://github.com/YunoHost/example_ynh/blob/15bf6f665e108a6086122158ae7d042660b66626/scripts/install#L44

The other solution is to define the env var in python, and modify it in _common.sh if needed. Does it solve all our problems ?

maniackcrudelis:
To define it into the core could as well be a solution indeed.
Or I thought about that solution

Into the php helper script:
```
YNH_PHP_VERSION=${YNH_PHP_VERSION:-$YNH_DEFAULT_PHP_VERSION}
```
Into the _common.sh of an app that need to specify an other version of php:
```
YNH_PHP_VERSION=7.4
```
That way, $YNH_PHP_VERSION will be set whether by the app, or by the helper. One way or the other, the variable will be available for the entire script from the beginning.

kay0u:
This solution looks good to me! That way we don't have to source _common.sh after official helpers, php and nginx helpers are both happy and we don't have to modify all our apps!

maniackcrudelis:
Considering the upgrade of this PR, I realized that we have to set YNH_DEFAULT_PHP_VERSION as an env variable from the core directly, because we use the variable into https://github.com/YunoHost/example_ynh/pull/110/files#diff-dc2801cb46a5d60e62234cff98084498R8, which will be defined before the source of the official helpers.
I'll update my other PR

maniackcrudelis:
I updated this PR, so now
The nginx and php-fpm configs will use the default php version.
Only the backup and restore scripts will use the version that has been used by the app.

A packager willing to use a custom version of php will set `YNH_PHP_VERSION=7.4` into his `_common.sh` and use `$YNH_PHP_VERSION` instead of `$YNH_DEFAULT_PHP_VERSION` into the scripts.
I'll make the PR to support this on the helper side.

scripts/backup
```diff
@@ -31,6 +31,7 @@ app=$YNH_APP_INSTANCE_NAME
 final_path=$(ynh_app_setting_get --app=$app --key=final_path)
 domain=$(ynh_app_setting_get --app=$app --key=domain)
 db_name=$(ynh_app_setting_get --app=$app --key=db_name)
+phpversion=$(ynh_app_setting_get --app=$app --key=phpversion)
```
kay0u: Uuh, where is set this value?

scripts/_common.sh
```diff
@@ -5,7 +5,7 @@
 #=================================================
 
 # dependencies used by the app
-pkg_dependencies="deb1 deb2"
+pkg_dependencies="deb1 deb2 php$YNH_DEFAULT_PHP_VERSION-deb1 php$YNH_DEFAULT_PHP_VERSION-deb2"
```
kay0u: `$YNH_DEFAULT_PHP_VERSION` or `$YNH_PHP_VERSION` ?

maniackcrudelis: Unless the packager set `$YNH_PHP_VERSION` to use a specific version, the variable is not yet set. It will be only when the official php helper is sourced.
So `$YNH_DEFAULT_PHP_VERSION` in most cases.

PR #114, opened 2020-04-28
fix typos...

ericgaspar:


maniackcrudelis:
Well no.
A quote is a quote. And a simple regular one ' is far better.

ericgaspar:
The devil is in the detail. [an explanation here](https://www.brunobernard.com/des-textes-plus-elegants-avec-les-apostrophes-typographiques/).

 And a simple regular one ' may be better for code but not for regular text.

maniackcrudelis:
This is a piece of code, we're on GitHub.
And the simple regular quote is the quote everyone got with his keyboard.

Considering the 2 first paragraph and the conclusion (to know how to do differently is of no use for me), the answer to the "question"
> J’espère néanmoins vous avoir convaincu

is, absolutely not.
To be honest, that kind of thing is for me just useless things to consider when you really have nothing else to do.

A quote is a quote. Everyone read and understand that as a quote. Human and machine.

ericgaspar:
I don't agree, mais bon...

PR #116, opened 2020-05-07
Clarify some things about backup script

alexAubin:
## Problem

Some weird things happening in backup script
- `ynh_backup` only declare files to be backuped and does not perform any "real" copy whatsoever
- Hence it's overkill and misleading to have progress bars ... In fact on a real server if you launch a "backup create", you will see your N apps running the backup scripts super quick and flooding messages with progress bar as if so much stuff happened but nope ... The real stuff happens later when the archive is actually created with the files declared by `ynh_backup`
- And apart from maybe specific edge cases, there's no need to start/stop the service during backup operations ... Start/stopping the service creates a service interruption that may disrupt experience for users...
- One could think that the service should be stopped during mysql dump, but nope, the helper uses a  --single-transaction ... So unless you're doing something very specific, no need to start/stop the service

## Solution

- Tweak the backup script, have 2 steps. One about declaring files. The second one about creating database dumps. There's no point in having progress bars.

## PR Status
- [X] Code finished.

### DIFF
--- a/scripts/backup
+++ b/scripts/backup
@@ -24,7 +24,7 @@ ynh_abort_if_errors
 #=================================================
 # LOAD SETTINGS
 #=================================================
-ynh_script_progression --message="Loading installation settings..." --time --weight=1
+ynh_print_info --message="Loading installation settings..."
 
 app=$YNH_APP_INSTANCE_NAME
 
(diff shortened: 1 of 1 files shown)


alexAubin:
Any opinion on this ? :/

Tagadda:
LGTM.
Stoping and starting the service could be commented instead of deleted, as some application could use that, but I'm okay with both.

kay0u:
> LGTM.
> Stoping and starting the service could be commented instead of deleted, as some application could use that, but I'm okay with both.

I guess, I don't know how many applications have to stop/restart their service for a backup.

We have the same type of mechanism for change_url: https://github.com/YunoHost/example_ynh/blob/c07802d4effb01019a64a801d96383946b2a35d0/scripts/change_url#L77

Tagadda:
> > LGTM.
> > Stoping and starting the service could be commented instead of deleted, as some application could use that, but I'm okay with both.
> 
> I guess, I don't know how many applications have to stop/restart their service for a backup.
> 
> We have the same type of mechanism for change_url:
> 
> https://github.com/YunoHost/example_ynh/blob/c07802d4effb01019a64a801d96383946b2a35d0/scripts/change_url#L77

For change_url, some apps need to know the path, so a restart may be useful.

alexAubin:
Will merge in 3 days if that's okay then

scripts/backup
```diff
@@ -33,46 +33,37 @@ domain=$(ynh_app_setting_get --app=$app --key=domain)
...
 #=================================================
-# STANDARD BACKUP STEPS
+# DECLARE DATA AND CONF FILES TO BACKUP
 #=================================================
-# STOP SYSTEMD SERVICE
-#=================================================
-ynh_script_progression --message="Stopping a systemd service..." --time --weight=1
 
-ynh_systemd_action --service_name=$app --action="stop" --log_path="/var/log/$app/$app.log"
+ynh_print_info --message="Declaring files to be backuped..."
```
JimboJoe, line 39:
```suggestion
ynh_print_info --message="Declaring files to be backed up..."
```

PR #127, opened 2020-06-04
Service integration

yalh76:
## Problem
- *Service integration steps are not homogenization*

## Solution
- *Current service integration steps homogenization*
- *adding the service integration step during upgrade*


yalh76:
> 
> 
> I would use "integrate **into** YunoHost" instead of "in".

no opinion on that point.

I just tried to homogenize between different scripts

PR #134, opened 2020-10-09
Add port for ynh_add_nginx_config

Krakinou:
## Problem
- When ` __ PORT__` is used in nginx.conf, ynh_add_nginx_config will not changed it if $port is not initialized with the port number. This will cause nginx to fail at reload.

## Solution
- *Initialize $port from the app settings. If port is not set in app settings, nothing will happen*

## PR Status
- [ ] Code finished.

### DIFF
--- a/scripts/change_url
+++ b/scripts/change_url
@@ -28,6 +28,7 @@ ynh_script_progression --message="Loading installation settings..." --time --wei
 
 # Needed for helper "ynh_add_nginx_config"
 final_path=$(ynh_app_setting_get --app=$app --key=final_path)
+port=$(ynh_app_setting_get --app=$app --key=port)
 
 # Add settings here as needed by your application
 #db_name=$(ynh_app_setting_get --app=$app --key=db_name)


yalh76:
There is no `__PORT__` in the nginx.conf provided in example_ynh, so that could be strange for a new developper... 
In addition, in that case `$port` initialisation is missing in the upgrade script.

For me, example_ynh should be as generical as possible without implementing all possibilities but providing a good package start and main guidelines

ericgaspar:
@Krakinou as a point. It is confusing not to mention $port variable... this variable should at least appear as a commented option. (like the database variables L.32-35) I encountered several time errors in the change_url script before realizing that the port variable was missing ...

alexAubin:
(As a followup, related to https://github.com/YunoHost/yunohost/pull/1118 , maybe `ynh_add_config` could report an error if it found some __FOOBAR__ to be replaced but no corresponding variable to replace it with)

yalh76:
Now `ynh_add_config` report an error when an argument s missing :)

yalh76:
Now `ynh_add_config` report an error when an argument is missing :)

PR #158, opened 2021-04-24
Set specific path for en or fr doc

ericgaspar:
Set specific path for en or fr doc


alexAubin:
Shouldn't we just point to https://yunohost.org/install instead ? It automagically redirects to the appropriate language (in fact, same thing for app docs, such as https://yunohost.org/app_wordpress)

yalh76 (approved):

PR #162, opened 2021-05-12
Some bashification of scripts

Salamandar:
Some change suggestions in the bash scripts.

* Use of `#!/usr/bin/env bash` instead of `#!/usr/bin/bash` for the scripts shebangs. 
  * It shouldn't change a thing but may be a good practice to be compatible with environments where bash would be at /bin/bash and not /usr/bin/bash.
* Use of an array for pkg_dependencies : that allows to split the declaration across multiple lines.
* Harmonization of the use of double quotes. One could say $app doesn't need double quotes, but it is quoted at some places already. So, well, it doesn't cost much and may prevent some issues.

### DIFF
--- a/scripts/_common.sh
+++ b/scripts/_common.sh
@@ -1,11 +1,15 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 #=================================================
 # COMMON VARIABLES
 #=================================================
 
 # dependencies used by the app
-pkg_dependencies="deb1 deb2 php$YNH_DEFAULT_PHP_VERSION-deb1 php$YNH_DEFAULT_PHP_VERSION-deb2"
+pkg_dependencies=(
+    deb1 deb2
+    "php$YNH_DEFAULT_PHP_VERSION-deb1"
+    "php$YNH_DEFAULT_PHP_VERSION-deb2"
+)
 
 #=================================================
 # PERSONAL HELPERS
--- a/scripts/backup
+++ b/scripts/backup
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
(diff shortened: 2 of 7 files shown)


yalh76:
> 
> 
> Some change suggestions in the bash scripts.
> 
>     * Use of `#!/usr/bin/env bash` instead of `#!/usr/bin/bash` for the scripts shebangs.
>       
>       * It shouldn't change a thing but may be a good practice to be compatible with environments where bash would be at /bin/bash and not /usr/bin/bash.

But for Debian, for now bash if at `/bin/bash` so no need to be compatible with environnements where bash would be in other place



> 
>     * Use of an array for pkg_dependencies : that allows to split the declaration across multiple lines.

It's already possible to have a multi line declaration: 
```
pkg_dependencies=" \
deb1 deb2 \
php$YNH_DEFAULT_PHP_VERSION-deb1 \
php$YNH_DEFAULT_PHP_VERSION-deb2 \
"
```

> 
>     * Harmonization of the use of double quotes. One could say $app doesn't need double quotes, but it is quoted at some places already. So, well, it doesn't cost much and may prevent some issues.

Yes it would be great but pushing that kind of changes will make packagers start a lot of PR to comply with example_ynh=>a lot of volunteer time for a small improvement... It should be better to start thinking of a new package format => https://github.com/YunoHost/example_ynh/pull/164#issuecomment-854957872

andretheolauret:
This PR looks inactive and non relevant ? Is useful to let him open ?

zamentur:
I agree with yalh76, the gain/time ratio is too bad and a lot of those changes will be fixed by the new format packaging.

I close the PR. 

Thank you for your contribution.

PR #164, opened 2021-05-28
Move repeated code into functions in _common.sh

Jules-Bertholet:
## Problem

Most of the scripts contain lots of duplicate code that is shared with other scripts. Maintaining multiple copies of the same code is unnecessarily time-consuming and error-prone.

## Solution

This PR moves repeated code blocks into bash functions in `_common.sh`, ensuring that there is only one copy that needs to be modified/updated.

## PR Status

### DIFF
--- a/scripts/_common.sh
+++ b/scripts/_common.sh
@@ -11,6 +11,146 @@ pkg_dependencies="deb1 deb2 php$YNH_DEFAULT_PHP_VERSION-deb1 php$YNH_DEFAULT_PHP
 # PERSONAL HELPERS
 #=================================================
 
+function test_folder {
+	test ! -e "$final_path" || ynh_die --message="$final_path path already contains a directory"
+}
+
+function install_dependencies {
+	### `ynh_install_app_dependencies` allows you to add any "apt" dependencies to the package.
+	### Those deb packages will be installed as dependencies of this package.
+	### If you're not using this helper:
+	###		- Remove the section "REMOVE DEPENDENCIES" in the remove script
+	###		- Remove the variable "pkg_dependencies" in _common.sh
+	###		- As well as the section "REINSTALL DEPENDENCIES" in the restore script
+	###		- And the section "UPGRADE DEPENDENCIES" in the upgrade script
+
+	ynh_install_app_dependencies $pkg_dependencies
+}
+
+function configure_system_user {
+		ynh_system_user_create --username=$app --home_dir="$final_path"
+}
+
+function setup_source {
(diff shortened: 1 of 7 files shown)


tituspijean (commented):
While this is *very* tempting, I think adding another layer of abstraction on top of the helpers is not suitable. The example app is meant as teaching method to showcase how app packaging can be done. If you move/hide the helpers and their parameters in another file, you hinder the reading of the scripts.

Every app is different. Implementing blanket functions hinders code flexibility.

*However*, what you are suggesting may be the base for a discussion on how to improve the packaging and app handling altogether. 

- I cannot count how many times I altered one script and forgot to do the same alteration in the others.
- A function to load all settings totally makes sense to me.
- etc.

alexAubin:
To elaborate a bit more on this topic : 
- I think nowadays there's a general consensus on the fact that the packaging format ain't super great, all 5 major scripts are a symmetric of each other and it's boring/hell to maintain all of them in sync
- Yet we kinda arrived to a good and stable recommended set of standard practices reflected in the example_ynh app ... and this goes includes *not* factorizing common stuff into _common.sh (apart from stuff like list of apt dependencies or custom/future helpers) ... even though not repetition is blasphemy from a dev's point of view ... and I personally was against not factorizing in the beginning, but "having a recommended set of standard practices" is definitely the most important part here, and refactorizing all 200+ apps brings little benefit
- With that said, there's an ongoing discussion since a while now about creating a new declarative packaging format which would address the current issue, and unlock many possibilities in terms of UX and other integration stuff with Yunohost, while ideally increasing the robustness and lowering the maintenance cost on the long term 
    - there are some thoughts here https://github.com/YunoHost/issues/issues/1606 and here https://github.com/YunoHost/issues/issues/1614 and some proofs of concepts, but it's kind of in stand by ... though considering Buster is progressing well, maybe i'll try to continue the work on this during summer

Jules-Bertholet:
In my personal experience, as soon as things get complicated (e.g., need to compile frontend and backend, each with their own toolchain that needs to be installed and managed, then set up 3 config files, 4 systemd services, 4 log files, etc), functions can quickly become essential to keep the complexity somewhat manageable.  But for simpler apps maybe they lead to useless complexity?

As for declarative packaging: there's always containers I guess… but that would mean rewriting a lot of stuff

zamentur:
Some work are made in core to solve this issue https://github.com/YunoHost/yunohost/pull/1289

tituspijean:
Meanwhile, let's close. :)

PR #166, opened 2021-07-06
Hardened password

ericgaspar:
- *Fix cases where an app needs a password longer than 4 characters, which can fail the CI (for example in phpBB)*

### DIFF
--- a/check_process
+++ b/check_process
@@ -10,7 +10,7 @@
 		admin="john"
 		language="fr"
 		is_public=1
-		password="strong-password"
+		password="1Strong-Password"
 		port="666"
 	; Checks
 		pkg_linter=1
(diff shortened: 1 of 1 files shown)


check_process
```diff
@@ -10,7 +10,7 @@
 		admin="john"
 		language="fr"
 		is_public=1
-		password="pass"
+		password="strong-password"
```
yalh76, line 13:
```suggestion
		password="1Strong-Password"
```
Maybe something more:  with a least one uppercase, a least one lowercase, at least one number, at least one special character and 16 characters ?

yalh76 (changes requested):

yalh76 (approved):

PR #168, opened 2021-08-19
Add purge option

ericgaspar:
- Add --purge option to "yunohost app remove" ([#1285](https://github.com/YunoHost/yunohost/pull/1285))

To be merged after version 4.2.8 is released.

### DIFF
--- a/scripts/remove
+++ b/scripts/remove
@@ -80,7 +80,7 @@ ynh_secure_remove --file="$final_path"
 # REMOVE DATA DIR
 #=================================================
 
-# Remove the data directory if --purge option is used.
+# Remove the data directory if --purge option is used
 if [ "$YNH_APP_PURGE" == true ]
 then
 	ynh_script_progression --message="Removing app data directory..." --time --weight=1
--- a/scripts/remove
+++ b/scripts/remove
@@ -81,7 +81,7 @@ ynh_secure_remove --file="$final_path"
 #=================================================
 
 # Remove the data directory if --purge option is used
-if [ "$YNH_APP_PURGE" == true ]
+if [ "$YNH_APP_PURGE" == 1 ]
 then
 	ynh_script_progression --message="Removing app data directory..." --time --weight=1
 	ynh_secure_remove --file="$datadir"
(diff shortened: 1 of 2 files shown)


scripts/remove
```diff
@@ -75,6 +76,17 @@ ynh_script_progression --message="Removing app main directory..." --time --weigh
 # Remove the app directory securely
 ynh_secure_remove --file="$final_path"
 
+#=================================================
+# REMOVE DATA DIR
+#=================================================
+
+# Remove the data directory if --purge option is used
+if [ "$YNH_APP_PURGE" == true ]
```
alexAubin, line 84:
```suggestion
if [ "$YNH_APP_PURGE" == True ]
```

It's gonna be True with a caps (because Python™), but we could also decide to go with 0 / 1 like it's done for other vars, i dunno

kay0u: I prefer to keep bash in a bash script :x

alexAubin: We can change it but the current implementation [here](https://github.com/YunoHost/yunohost/blob/dev/src/yunohost/app.py#L1236) is `str(some_bool)` which yields `True` of `False`

I don't have a strong opinion on what to do exactly, it's true that we could have a more consistent convention across all the helpers etc, 'cause indeed in some other case we have `if [[ $SOME_VAR == 0 ]]` (or 1) ...

zamentur (approved):

ericgaspar:
I am using the remove with purge optin in Galène script and I get this warning...

`Warning: ./remove: line 59: [: False: integer expression expected`

I guess we need to declare somewhere if `YNH_APP_PURGE` is true or false.

alexAubin:
Uuuuuh yeah as I was trying to point out in some earlier comment, because of the way it's defined [here](https://github.com/YunoHost/yunohost/blob/dev/src/yunohost/app.py#L1238), the value is gonna be `True` or `False`. If we want, we may still tweak the python code to yield 0 or 1 instead, if we want more consistency with other pieces of code where we check `if [[ $some_var -eq 1 ]]`

alexAubin:
To be fixed by https://github.com/YunoHost/yunohost/commit/ad3602a24fffe9fa16c0d767a1d151e0ea966de6 in next 4.3 iteration

PR #169, opened 2021-08-19
Add no_safety_backup option

ericgaspar:
- Add --no-safety-backup option to “yunohost app upgrade” [#1286](https://github.com/YunoHost/yunohost/pull/1286)

To be merged after version 4.2.8 is released.

### DIFF
--- a/scripts/upgrade
+++ b/scripts/upgrade
@@ -38,16 +38,19 @@ upgrade_type=$(ynh_check_app_version_changed)
 #=================================================
 # BACKUP BEFORE UPGRADE THEN ACTIVE TRAP
 #=================================================
-ynh_script_progression --message="Backing up the app before upgrading (may take a while)..." --time --weight=1
 
-# Backup the current version of the app
-ynh_backup_before_upgrade
-ynh_clean_setup () {
-	# Restore it if the upgrade fails
-	ynh_restore_upgradebackup
-}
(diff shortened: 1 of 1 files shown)


scripts/upgrade
```diff
@@ -38,16 +38,19 @@ upgrade_type=$(ynh_check_app_version_changed)
...
-ynh_clean_setup () {
-	# Restore it if the upgrade fails
-	ynh_restore_upgradebackup
-}
-# Exit if an error occurs during the execution of the script
-ynh_abort_if_errors
+
+# Backup the current version of the app unless --no_safety_backup option is used
+if [ "$NO_BACKUP_UPGRADE" -eq 0 ]
+then
```
alexAubin, line 44:
```suggestion
# Backup the current version of the app unless --no_safety_backup option is used
if [ "$NO_BACKUP_UPGRADE" -eq 0 ]
then
```

Uuuuh but the thing is that this if is already inside the `ynh_backup_before_upgrade` helper : https://github.com/YunoHost/yunohost/blob/dev/data/helpers.d/backup#L437 so there's no huge gain in adding the if in every upgrade script ... In fact on the long term, we should aim to *remove* this whole snippet and let this mecanism be handled in the core instead of copy-pasting the same 5~10 lines in every apps

ericgaspar:
bon ben on close

PR #172, opened 2021-09-04
Implement GitHub Action to automatically upgrade from upstream

tituspijean:
## Problem

- App maintenance is tedious and can be repetitive

## Solution

- Use GitHub Actions with a cron job to check daily for new releases from the upstream app. If a new release is detected:
   - `updater.sh` is run to update the source files (script to be edited by the maintainers, hopefully once and for all)
   - a PR is then created. 🎉 

N.B.:
- Automatic CI testing depends on: https://github.com/YunoHost/yunorunner/pull/25
- Action template is in https://github.com/YunoHost-Apps/.github

To do:
- [x] Per [this](https://github.com/YunoHost-Apps/nodered_ynh/runs/3522415806?check_suite_focus=true), we need to add more `if` conditions or whatever to stop the workflow when a PR is already opened.
- [x] Factorise the yml even more, with a generic Action for all apps. You can see the organization's workflow when adding new Actions: e.g. https://github.com/YunoHost-Apps/nodered_ynh/actions/new, "Workflows created by YunoHost-Apps"


yalh76 (approved):

.github/workflows/updater.sh
```diff
@@ -0,0 +1,132 @@
...
+
+#=================================================
+# GENERIC FINALIZATION
+#=================================================
+
+# Install moreutils, needed for sponge
+sudo apt-get install moreutils
+
+# Replace new version in manifest
+jq -s --indent 4 ".[] | .version = \"$VERSION~ynh1\"" manifest.json | sponge manifest.json
```
ericgaspar, line 126:
```suggestion
jq -s --indent 4 ".[] | .version = \"$version~ynh1\"" manifest.json | sponge manifest.json
```

ericgaspar (changes requested):

.github/workflows/updater.sh
ericgaspar: Following the chat debug session :)

.github/workflows/updater.sh
```diff
@@ -0,0 +1,129 @@
...
+# FETCHING LATEST RELEASE AND ITS ASSETS
+#=================================================
+
+# Fetching information
+current_version=$(cat manifest.json | jq -j '.version|split("~")[0]')
+repo=$(cat manifest.json | jq -j '.upstream.code|split("https://github.com/")[1]')
+# Some jq magic is needed, because the latest upstream release is not always the latest version (e.g. security patches for older versions)
+version=$(curl --silent "https://api.github.com/repos/$repo/releases" | jq -r '.[] | select( .prerelease != true ) | .tag_name' | sort -V | tail -1)
+assets=($(curl --silent "https://api.github.com/repos/$repo/releases" | jq -r '[ .[] | select(.tag_name=="'$version'").assets[].browser_download_url ] | join(" ") | @sh' | tr -d "'"))
+
```
ericgaspar, line 25:
```
if [[ ${version:0:1} == "v" || ${version:0:1} == "V" ]]; then
    version=${version:1}
fi
```
Is it worth adding this little chunk?

tituspijean: There may be plenty other ways for tag names not to fit quite well, but I guess this one may occur often. Let's add it yes. 😃

ericgaspar (approved):

PR #179, opened 2022-01-02
WIP: Manifest/packaging v2

alexAubin:
## Problem

## Solution

## PR Status

- [ ] Code finished and ready to be reviewed/tested

## Automatic tests


ericgaspar:
Will packaging v2 allow parametric menus (I don't know if this is the correct term, the idea being that if we select a database to be installed (for my_webapp for example) then we get the database settings option.)

alexAubin:
You mean conditional questions during install ? I have nothing in mind planned for this ... i think it may or may not already work as an indirect consequence of the work on config panel, @zamentur would know more about this ;)

Salamandar:
I'm currently doing a WIP that basically generates the install,remove etc scripts from a config file.
I didn't see this PR, so I did something different on my own, but I'm sure our experiments will graciously merge over time 🥰

scripts/upgrade
```diff
@@ -118,48 +73,37 @@ then
...
-#=================================================
+ynh_add_systemd_config
+
+yunohost service add $app --description="A short description of the app" --log="/var/log/$app/$app.log"
+
+ynh_use_logrotate --non-append
+
+ynh_add_fail2ban_config --logpath="/var/log/nginx/${domain}-error.log" --failregex="Regex to match into the log for a failed login"
 
+
```
alexAubin, line 104: So here I went for grouping all the "system conf related to this app" instructions because imho it's much more sensible ... (and similar stuff in the install and restore script) ... all of these should take pretty much no time and imho it's unnecessarily complex both from a dev and user point of view to split those into several micro-steps ... Instead, this pave the way for packaging v3 (yeah :sweat_smile: ) where these are expected to be handled at the same time via some declarative paradigm such that we end up with something like : 

- (by the core) App resources are provisioned : system user, install/data dir, apt dependencies, port, database, ...
- (by the app script) App is built : fetch sources, run whichever pip install or npm build ... ideally as non-root
- (by the core) System configs related to the app are deployed
- (by the app script) Run app initial configuration or DB migration for the app
- (by the core) ??? Restart app services maybe

PR #183, opened 2022-02-08
Update upgrade

ericgaspar:
Following Alex fix for Matomo: https://github.com/YunoHost-Apps/matomo_ynh/pull/69/commits/aa41db0fd0b7127617d30fdce5e7811079ab18d0

When we change PHP version PHP7.3 -> PHP8.0. We need to download the dependencies before updating the app.

### DIFF
--- a/scripts/upgrade
+++ b/scripts/upgrade
@@ -102,6 +102,13 @@ if ! ynh_permission_exists --permission="api"; then
 fi
 
 #=================================================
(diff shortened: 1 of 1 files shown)


yalh76:
I disagree...

@alexAubin made that change because the "DOWNLOAD, CHECK AND UNPACK SOURCE" step do more than that ..
It also do a `console core:update -n` that should be in the `SPECIFIC UPGRADE` step

PR #190, opened 2022-06-06
Add License as default

ericgaspar:
- Add license GNU AFFERO GENERAL PUBLIC LICENSE Version 3 as default

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,661 @@
-File containing the license of your package.
+GNU AFFERO GENERAL PUBLIC LICENSE
+   Version 3, 19 November 2007
 
-More information here:
-https://yunohost.org/packaging_apps_guidelines#yep-1-3
+Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+Everyone is permitted to copy and distribute verbatim copies
+of this license document, but changing it is not allowed.
+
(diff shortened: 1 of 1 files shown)


ericgaspar:
May I ask why so much confusion?  Is GNU AFFERO GENERAL PUBLIC LICENSE Version 3 not the default for YunoHost package? packager can always put something else. This is a guideline.

yalh76:
as far as I know, there is no default licence for YunoHost packages

ericgaspar:
> as far as I know, there is no default licence for YunoHost packages

This PR adresses the lack default licence. A very large majority of licence are AGPL-3.0, making It the _the facto_ licence.

yalh76:
> > as far as I know, there is no default licence for YunoHost packages
> 
> This PR addresses the lack of default licence. A very large majority of licence are AGPL-3.0, making It the _the facto_ licence.

Do you some statistics about that? like for each licence used by packages, the number of time that licence is used ?

ericgaspar:
I do not have proper metrics (as per specific percentage)
The license file is confusing because you don't know which license to use... Sometimes packagers put upstream app licenses. 
[wemawema_ynh](https://github.com/YunoHost-Apps/wemawema_ynh) is intentionally using a specific license

PR #198, opened 2023-01-02
Create application log directory in restore

gcollin:
## Problem
When restoring an application based on this example, you have to re-create the /var/log application directory in case it has been deleted.
This happens if the application have been backup, removed and then restored. 

## Solution

Added a mkdir in the restore script

## PR Status

- [x] The fix/enhancement were manually tested (if applicable)

## Automatic tests
Not needed as it's just an example script

### DIFF
--- a/scripts/restore
+++ b/scripts/restore
@@ -155,6 +155,7 @@ systemctl enable $app.service --quiet
 ynh_script_progression --message="Restoring the logrotate configuration..." --weight=1
 
 ynh_restore_file --origin_path="/etc/logrotate.d/$app"
+mkdir --parents /var/log/$app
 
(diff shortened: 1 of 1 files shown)


Salamandar:
This should actually be done at install too.

Tagadda (approved):

OniriCorpe:
Maybe it will be better if the core handle this with the v2 packaging?

So the core would create `"/var/log/$app"` at install or at upgrade if it doesn't exist

Salamandar:
Yes, this is actually in the packaging v3 "wish list"

Salamandar:
It should be closed now that #217  is merged

PR #201, opened 2023-02-20
Simplify change_url syntax

alexAubin:
Change URL script is hell, simplified via changes introduced in 11.1.8 (9?) : https://github.com/YunoHost/yunohost/commit/2b70ccbf40c6143fb1bfdd7d7414864a897d9542

In particular:
- `new`/`old_domain`, `new`/`old_path`, `change_domain`/`path` are now automatically available, no need for the boing computation anymore
- new helper `ynh_change_url_nginx_config` handles the part about tweaking the nginx config
- I suggest to just not backup/restore the app anymore because this is way overkill, in most cases it's only going to be updating the nginx conf (which is rolled back by the core if for some reason this fails) and updating the app's conf, which is unlikely to fail ... and the app that really do heavy stuff can still have a safety backup mechanism ...

### DIFF
--- a/scripts/change_url
+++ b/scripts/change_url
@@ -10,64 +10,6 @@ source _common.sh
 source /usr/share/yunohost/helpers
 
 #=================================================
-# RETRIEVE ARGUMENTS
-#=================================================
-
-old_domain=$YNH_APP_OLD_DOMAIN
-old_path=$YNH_APP_OLD_PATH
-
-new_domain=$YNH_APP_NEW_DOMAIN
-new_path=$YNH_APP_NEW_PATH
-
-app=$YNH_APP_INSTANCE_NAME
-
-#=================================================
-# LOAD SETTINGS
-#=================================================
-ynh_script_progression --message="Loading installation settings..." --weight=1
-
-# Needed for helper "ynh_add_nginx_config"
-final_path=$(ynh_app_setting_get --app=$app --key=final_path)
-
-# Add settings here as needed by your application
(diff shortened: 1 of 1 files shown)


scripts/change_url
```diff
@@ -120,13 +40,6 @@ ynh_script_progression --message="Starting a systemd service..." --weight=1
 
 ynh_systemd_action --service_name=$app --action="start" --log_path="/var/log/$app/$app.log"
```
zamentur, line 42: i think nginx reload should be done after service restart... to avoid 502

alexAubin: Is this really a thing ........ I've seen countless situation where a service is down / restarted / whatever, and just restarting the service is enough to get it back online, there's no need to reload nginx after restarting a service ...

alexAubin: And anyway the core already does reload nginx after the script is ran ... https://github.com/YunoHost/yunohost/blob/dev/src/app.py#L528

zamentur: "Is this really a thing ........" on my side, i have seen several webapp where it's important to do it... I don't know excatly why, but be sure that on all our app, it's needed.

However it's fine for the reload by the core, so it's fixed, sorry for the noise.

alexAubin: It would still be quite interesting to have a way to reproduce the issue you are mentioning to know once and for all wether or not this is a thing ... It has other implications for other packaging parts designs

yalh76 (approved):

PR #208, opened 2023-06-08
extra_php-fpm.conf

ericgaspar:
Set extra_php-fpm.conf


alexAubin (commented):
Eeeeeh this is a good change though bro

Just ideally would like to improve the situation with the management of the fpm_usage/footprint setting but otherwise we clearly should push for the extra-fpm conf usage rather than the old wall of php conf filled with 90% comments

PR #216, opened 2023-11-05
Create log files at install and restore steps

Salamandar:
## Problem

fail2ban fails if the log file does not exist yet

## Solution

Create the file at install and restore steps.

## PR Status

## Automatic tests

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -26,7 +26,7 @@ source /usr/share/yunohost/helpers
 # ...
 
 #
-# $app is the app id (i.e. 'example' for first install, 
+# $app is the app id (i.e. 'example' for first install,
 # or 'example__2', '__3', ... for multi-instance installs)
 #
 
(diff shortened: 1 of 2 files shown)


OniriCorpe:
Not necessary if this PR is accepted: https://github.com/YunoHost/yunohost/pull/1737

Salamandar:
> Not necessary if this PR is accepted: [YunoHost/yunohost#1737](https://github.com/YunoHost/yunohost/pull/1737)

Partially, as the log directory containing the touched file won’t be created by your merge request. Also, I added a comment about permissions.

Though, I would prefer to fix this issue directly in the helpers, as you did in your merge request :)

scripts/install
```diff
@@ -96,6 +96,12 @@ ynh_add_systemd_config
 
 yunohost service add $app --description="A short description of the app" --log="/var/log/$app/$app.log"
 
+# Create the log file (required for fail2ban)
+mkdir -p "/var/log/$app"
+touch "/var/log/$app/$app.log"
+chown -R "$app:$app" "/var/log/$app"
+chmod -R u=rwX,g=rX,o= "/var/log/$app"
```
alexAubin, line 103: I think that one is now handled by the logrotate helper ... Assuming the app calls the logrotate helper ... 

But it's not done in the restore script, because during the restore script we restore the raw backup file instead of calling a symetric helper, which imho is bullshit but hmpf

Salamandar: Indeed. But the real issues happens when you're using fail2ban but not logrotate (even though you just should use logrotate)… Or *before* logrotate. 

Anyways until packaging v3 maybe I'll just update this merge request to only mkdir in the restore script, do you agree ?

Salamandar:
Closing in favor of #217

PR #217, opened 2023-11-05
Backup and restore the log directory

Salamandar:
## Problem

Log directory is not backed up

## Solution

well…

Also this is a bit a concurrent of https://github.com/YunoHost/example_ynh/pull/216

## PR Status

## Automatic tests

### DIFF
--- a/scripts/backup
+++ b/scripts/backup
@@ -62,6 +62,10 @@ ynh_backup --src_path="/etc/cron.d/$app"
 
 ynh_backup --src_path="/etc/$app/"
 
+# For apps with huge logs, you might want to pass --is_big,
+# and in restore script, mkdir and pass --not_mandatory to ynh_restore_file.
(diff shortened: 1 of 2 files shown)


scripts/backup
```diff
@@ -74,6 +74,8 @@ ynh_backup --src_path="/etc/cron.d/$app"
 
 ynh_backup --src_path="/etc/$app/"
 
+ynh_backup --src_path="/var/log/$app/"
```
alexAubin, line 77: Hmmmmmm 95% yes but then I'm wondering about for example synapse for which logs have an absolutely unreasonable size ... but that's probably solved by saying this is just a recommendation for the nominal use case and synapse can just not implement it, or add --is_big

alexAubin (approved):

scripts/backup
Salamandar: Indeed !
I'll add a comment.

PR #225, opened 2024-01-10
On upgrade, pass --full_replace, --keep to ynh_setup_source, with the according doc.

Salamandar:


scripts/upgrade
```diff
@@ -69,8 +69,13 @@ ynh_systemd_action --service_name="$app" --action="stop" --log_path="/var/log/$a
...
     ynh_script_progression --message="Upgrading source files..." --weight=1
 
+    ### ynh_setup_source can wipe the destination dir if called with --full_replace.
+    ### On upgrade, that is certainly what you want, to remove any old source file that
+    ### does not exist in the new version of the software.
+    ### You can list with --keep every file/directory to *not* wipe or overwrite,
+    ### useful for configuration files, data directories, or plugins.
     # Download, check integrity, uncompress and patch the source from manifest.toml
-    ynh_setup_source --dest_dir="$install_dir"
+    ynh_setup_source --dest_dir="$install_dir" --full_replace=1 --keep=".env data"
```
alexAubin, line 78: On the one hand, yes

On the other hand, people should reaaaaaaaaally be careful with this because if we yolowrite this without checking what is the real usage of every folder, we may straight up delete real-life data that some app may store in their install dir ... On the other hand, the install dir should always be backed up by the backup dir ... so idk

PR #226, opened 2024-01-11
Rework upgrade, remove the 'if upgrade_type == UPGRADE_APP' test.

Salamandar:
Indeed, even if the upgrade is only a 'ynh revision', this revision might actually be fixing what *is* inside this if test.
So we can't discriminate generic things in the script according to the upgrade_type.

The ynh_check_app_version_changed helper is now a simple shim that reads YNH_APP_UPGRADE_TYPE. The rest of its old logic is now handled by the core. So instead of calling it, we can just read YNH_APP_UPGRADE_TYPE and add some documentation about possible values.

### DIFF
--- a/scripts/upgrade
+++ b/scripts/upgrade
@@ -22,13 +22,12 @@ source /usr/share/yunohost/helpers
 ### - resources are automatically provisioned / updated / deleted (depending on existing resources)
 ### - a safety backup is automatically created by the core and will be restored if the upgrade fails
 
-### This helper will compare the version of the currently installed app and the version of the upstream package.
-### $upgrade_type can have 2 different values
-### - UPGRADE_APP if the upstream app version has changed
+### This variable describes which upgrade type is occurring, allowing the script to handle different modes:
 ### - UPGRADE_PACKAGE if only the YunoHost package has changed
-### ynh_check_app_version_changed will stop the upgrade if the app is up to date.
-### UPGRADE_APP should be used to upgrade the core app only if there's an upgrade to do.
-upgrade_type=$(ynh_check_app_version_changed)
+### - UPGRADE_APP if the upstream app version has changed
+### If your package needs to handle other things, like same-version upgrades or downgrades, please
+### check out the $YNH_APP_UPGRADE_TYPE variable that can contain DOWNGRADE and UPGRADE_SAME too.
+# upgrade_type=$(ynh_check_app_version_changed)
 
 #=================================================
 # STANDARD UPGRADE STEPS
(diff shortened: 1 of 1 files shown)


zamentur:
You miss a part of the code: https://github.com/YunoHost/yunohost/blob/6fdcf03e9200f25431e998b030602c74c2adbae8/helpers/utils#L998

So UPGRADE_FULL is sadly transformed into UPGRADE_APP

Salamandar:
> You miss a part of the code: https://github.com/YunoHost/yunohost/blob/6fdcf03e9200f25431e998b030602c74c2adbae8/helpers/utils#L998

No, for now i'm focusing on this variable :) IMHO the days of this helper are counted.

> So UPGRADE_FULL is transformed into UPGRADE_APP.

Indeed. So one more reason to remove UPGRADE_FULL.
> 
> If we removed the helper, I guess packagers could create bug by missing the UPGRADE_FULL usecase. I studied month ago this values on all apps, and only 1 apps was using UPGRADE_FULL (by mistake).

I fully agree.

scripts/upgrade
```diff
@@ -22,13 +22,13 @@ source /usr/share/yunohost/helpers
...
+### This variable describes which upgrade type is occurring, allowing the script to handle different modes:
+### - DOWNGRADE_FORCED if a downgrade is happening
+### - UPGRADE_FORCED if the version does not change
 ### - UPGRADE_PACKAGE if only the YunoHost package has changed
-### ynh_check_app_version_changed will stop the upgrade if the app is up to date.
-### UPGRADE_APP should be used to upgrade the core app only if there's an upgrade to do.
-upgrade_type=$(ynh_check_app_version_changed)
+### - UPGRADE_APP if the upstream app version has changed
+### - UPGRADE_FULL if both the YunoHost package and the app version changed
+upgrade_type=$YNH_APP_UPGRADE_TYPE
```
alexAubin, line 31: not sure to understand what that variable is used for then ? x_X

alexAubin: No app is using this variable for anything except the same copypasta of the example app, honestly let's just remove the hell of this from example_ynh entirely ...

Salamandar: @alexAubin it's not used anymore, i just left it for reference, but I can comment it actually, to let the packager know it's optional.

Salamandar:
@zamentur I reverted the use of `upgrade_type=$(ynh_check_app_version_changed)`, just commented this time.

scripts/upgrade
Salamandar: Now it's commented.

PR #246, opened 2025-04-19
Set file group permission to www-data

oleole39:
I find confusing that current recommendation for file permissions to `$app:$app` when everything else in the script is chowned to `$app:www-data` by default. 
Although I may miss a specific reason for it?

Also @alexAubin for an app you had recommended me to rather use `www-data` as in this PR.


alexAubin:
Urghhh it sounds like terrible security practice to give rights to www-data on the app's config because www-data should theoretically only have access to static files such as HTML, CSS, JS, fonts, images, uploaded files etc ...

Also in practice chgrp-ing doesnt change anything because of the previous `chmod 400` (= 0 perm for group), though still chown-ing to www-data is confusing with the intention

oleole39:
Ok, thanks for explaining

PR #249, opened 2026-01-02
last NGINX patch

ericgaspar:
## Problem

## Solution

## PR Status

- [ ] Code finished and ready to be reviewed/tested

## Automatic tests

### DIFF
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -12,14 +12,9 @@ location __PATH__/ {
 
   try_files $uri $uri/ index.php;
   location ~ [^/]\.php(/|$) {
-    fastcgi_split_path_info ^(.+?\.php)(/.*)$;
     fastcgi_pass unix:/var/run/php/php__PHP_VERSION__-fpm-__APP__.sock;
 
-    fastcgi_index index.php;
-    include fastcgi_params;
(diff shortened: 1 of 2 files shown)


conf/nginx.conf
```diff
@@ -12,14 +12,9 @@ location __PATH__/ {
...
   location ~ [^/]\.php(/|$) {
-    fastcgi_split_path_info ^(.+?\.php)(/.*)$;
     fastcgi_pass unix:/var/run/php/php__PHP_VERSION__-fpm-__APP__.sock;
 
-    fastcgi_index index.php;
-    include fastcgi_params;
-    fastcgi_param REMOTE_USER $remote_user;
-    fastcgi_param PATH_INFO $fastcgi_path_info;
-    fastcgi_param SCRIPT_FILENAME $request_filename;
+    include fastcgi_params_with_auth;
```
Josue-T, line 17: I would add more info here to explain when to use the `fastcgi...with_auth` and the other other one without auth. Because here it's not explicit to the packager that it's not always this one. Also this one should be used when we need the SSO feature. Otherwise we should use the other one.

Josue-T (changes requested):
Thanks for the PR.

Also ideally we would also add an example for the proxy pass whish is used for all non php apps.

